PR #1, opened 2016-03-08
Add a health check endpoint convention.

jessemyers:
Allows all Flask services to provide the same health checking interface
(e.g. to load-balancers, monitors, and other diagnostics). Provides a
simple hook for health reporting of sub-components.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,15 @@
 # microcosm-flask
 
-Opinionated Flask services
+Opinionated Flask services.
+
+[![Circle CI](https://circleci.com/gh/globality-corp/microcosm-flask/tree/develop.svg?style=svg)](https://circleci.com/gh/globality-corp/microcosm-logging/tree/develop)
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -1,3 +1,15 @@
...
+
+
+## Conventions
+
+ - Every Flask service has a health check endpoint at `/api/health`
+
+
+## Configuration
+
+ - The object graph's `debug` and `testing` flags are propagated to the Flask application.
```
srt32: I'm not sure I understand what this line means. Is there a way to override them if need be (is there a valid need?)?

jessemyers: You can set `flask.debug` and `flask.testing` when you are developing or running (unit) tests. The object graph has meta data that knows these settings because the context that "I am developing" or "I am running tests" is generally useful.

This line just means that the Flask application will be created with this same context.

PR #2, opened 2016-03-09
Add generalized Flask error handling.

jessemyers:
Basics:
 - All of the basic HTTP status codes and user exceptions get the same error handler by default
 - Errors are encoded as JSON (even for non-20X statuses) and contain three canonical values:
    - A message
    - A retryable flag
    - A context dictionary

### DIFF
--- a/microcosm_flask/errors.py
+++ b/microcosm_flask/errors.py
@@ -1,0 +1,111 @@
+"""
+Generalized error handling.
+
+"""
+from logging import getLogger
+
+from flask import jsonify
+from werkzeug.exceptions import default_exceptions
+
+
+error_logger = getLogger("errors")
+
+
+def extract_status_code(error):
+    """
+    Extract an error code from a message.
+
+    """
+    try:
+        return error.code
+    except AttributeError:
+        try:
+            return error.status_code
+        except AttributeError:
+            try:
+                return error.errno
+            except AttributeError:
+                return 500
+
+
+def extract_error_message(error):
+    """
+    Extract a useful message from an error.
+
+    Prefer the description attribute, then the message attribute, then
+    the errors string conversion. In each case, fall back to the error class's
+    name in the event that the attribute value was set to a uselessly empty string.
+
+    """
+    try:
+        return error.description or error.__class__.__name__
+    except AttributeError:
+        try:
+            return str(error.message) or error.__class__.__name__
+        except AttributeError:
+            return str(error) or error.__class__.__name__
+
+
+def extract_context(error):
+    """
+    Extract extract context from an error.
+
+    Errors may (optionally) provide a context attribute which will be encoded
+    in the response.
+
+    """
+    return getattr(error, "context", {})
+
+
+def extract_retryable(error):
+    """
+    Extract a retryable status from an error.
+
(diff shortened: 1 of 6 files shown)


microcosm_flask/errors.py
```diff
@@ -0,0 +1,111 @@
...
+Generalized error handling.
+
+"""
+from logging import getLogger
+
+from flask import jsonify
+from werkzeug.exceptions import default_exceptions
+
+
+error_logger = getLogger("errors")
```
jessemyers: I'll probably switch the main Flask logger at some point, but I haven't tested that it's getting configured properly yet.

microcosm_flask/errors.py
```diff
@@ -0,0 +1,111 @@
...
+
+    """
+    try:
+        return error.code
+    except AttributeError:
+        try:
+            return error.status_code
+        except AttributeError:
+            try:
+                return error.errno
```
jessemyers: Note that this would cause *client* errors from `Requests` to be propagated outward. That will often not be the correct behavior semantically, but it's an ok default.

microcosm_flask/errors.py
```diff
@@ -0,0 +1,111 @@
...
+    Register error handlers.
+
+    """
+
+    # override all of the werkzeug HTTPExceptions
+    for code in default_exceptions.iterkeys():
+        graph.flask.error_handler_spec[None][code] = make_json_error
+
+    # register catch all for user exceptions
+    graph.flask.error_handler_spec[None][None] = [(Exception, make_json_error)]
```
jessemyers: We didn't have anything like this in our cookie cutters as far as I know. This handles any non-werkzeug exception

setup.py
```diff
@@ -26,8 +26,10 @@
     ],
     entry_points={
         "microcosm.factories": [
+            "error_handlers = microcosm_flask.errors:configure_error_handlers",
             "health = microcosm_flask.conventions.health:configure_health",
             "flask = microcosm_flask.factories:configure_flask",
+            "app = microcosm_flask.factories:configure_flask_app",
```
jessemyers: Small change here: I'm using the binding `flask` to refer to the plain Flask instance and `app` to refer to a configured instance that uses conventions. That gives the various conventions a target to configure that doesn't create a circular dependency.

microcosm_flask/errors.py
adamhadani: :+1:

microcosm_flask/errors.py
adamhadani: yep. one thought - might be OK to propagate by default if debug/testing is True (either as the returned status code or as part of some error context field in the error JSON response)

microcosm_flask/errors.py
adamhadani: :+1:  important so no more HTML errors..

adamhadani:
LGTM

microcosm_flask/tests/test_errors.py
```diff
@@ -0,0 +1,200 @@
...
+
+
+class MyConflictError(Exception):
+    code = 409
+    retryable = True
+
+    @property
+    def context(self):
+        return dict(
+            resource="Banana!",
```
srt32: :banana:

PR #3, opened 2016-03-09
Add a basic auth decorator for Flask application.

jessemyers:
This change also fixes up (and acts as a testbed) for setting the Flask app's config.

### DIFF
--- a/microcosm_flask/basic_auth.py
+++ b/microcosm_flask/basic_auth.py
@@ -1,0 +1,72 @@
+"""
+Basic Auth support.
+
+"""
+from base64 import b64encode
+
+from flask.ext.basicauth import BasicAuth
+from werkzeug.exceptions import Unauthorized
+
+from microcosm.api import defaults
+from microcosm_flask.errors import with_headers
+
+
+def encode_basic_auth(username, password):
+    """
+    Encode basic auth credentials.
+
+    """
+    return "Basic {}".format(
+        b64encode("{}:{}".format(
+            username,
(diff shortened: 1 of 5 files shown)


microcosm_flask/basic_auth.py
```diff
@@ -0,0 +1,80 @@
...
+
+        """
+        return password is not None and self.credentials.get(username, None) == password
+
+    def challenge(self):
+        """
+        Override challenge to raise an exception that will trigger regular error handling.
+
+        """
+        response = super(ConfigBasicAuth, self).challenge()
```
jessemyers: The default behavior for this library was to return a Flask `Response` with a 401 status code. That works, but won't exercise our error handler logic and return JSON responses.

I'm raising my own error and injecting the expected `WWW-Authenticate` header with this `with_headers` wrapper (and accompanying changes in the error handling).

microcosm_flask/factories.py
```diff
@@ -16,7 +16,12 @@ def configure_flask(graph):
     app.debug = graph.metadata.debug
     app.testing = graph.metadata.testing
 
-    # TODO: wire in the graph's configuration to Flask's
+    # copy in the graph's configuration for non-nested keys
```
jessemyers: Basically: some third-party libraries look in `app.config` for their config. We want to avoid that where possible, but do propagate top-level keys from our configuration. The `BasicAuth` library is a good test case for this feature since it loads its *realm* from the config.

PR #4, opened 2016-03-09
Add logging and audit configuration to Flask app

jessemyers:
Several key changes:
 - `microcosm-logging` is now a dependency; I think this is an OK amount of coupling because logging is fundamental
 - An audit logging decorator is now part of the default setup; this one varies a bit from what we've been using elsewhere

### DIFF
--- a/microcosm_flask/audit.py
+++ b/microcosm_flask/audit.py
@@ -1,0 +1,98 @@
+"""
+Audit log support for Flask routes.
+
+"""
+from functools import wraps
+from json import loads
+from traceback import format_exc
+
+from flask import current_app, request
+
+from microcosm_flask.errors import extract_status_code, extract_error_message
+
+
+def audit(func):
+    """
+    Record a Flask route function in the audit log.
+
+    Generates a JSON record in the Flask log for every request.
+
+    """
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        response = None
+
+        # always include these fields
+        audit_dict = dict(
+            operation=request.url_rule.rule,
+            func=func.__name__,
+            method=request.method,
+        )
+
+        # include request body on debug (if any)
+        if current_app.debug and request.get_json(force=True, silent=True):
+            audit_dict["request_body"] = request.get_json(force=True)
+
+        try:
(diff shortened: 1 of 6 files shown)


microcosm_flask/audit.py
```diff
@@ -0,0 +1,98 @@
...
+    Generates a JSON record in the Flask log for every request.
+
+    """
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        response = None
+
+        # always include these fields
+        audit_dict = dict(
+            operation=request.url_rule.rule,
```
jessemyers: In anticipation of the next round of changes, I'm standardizing the name of the endpoint as "operation".

microcosm_flask/audit.py
```diff
@@ -0,0 +1,98 @@
...
+            func=func.__name__,
+            method=request.method,
+        )
+
+        # include request body on debug (if any)
+        if current_app.debug and request.get_json(force=True, silent=True):
```
jessemyers: Request and response content are now only sent with debug enabled. It's too big of a privacy risk to log content otherwise.

microcosm_flask/audit.py
```diff
@@ -0,0 +1,98 @@
...
+            audit_dict["request_body"] = request.get_json(force=True)
+
+        try:
+            response = func(*args, **kwargs)
+        except Exception as error:
+            audit_dict.update(
+                success=False,
+                message=extract_error_message(error)[:2048],
```
jessemyers: We get to reuse the error handling parsing logic here. I chose to name this attribute `message` to match the error response (instead of `error` or similar).

microcosm_flask/audit.py
```diff
@@ -0,0 +1,98 @@
...
+
+    return wrapper
+
+
+def parse_response(response):
+    """
+    Parse a Flask response into a body and status code.
+
+    The returned value from a Flask view could be:
+        * a tuple of (response, status) or (response, status, headers)
```
jessemyers: Our previous code had the parse logic here wrong. It got the response/status in the wrong order.

PR #5, opened 2016-03-10
Add a convention around named, resolveable Operations

jessemyers:
In order to enable HAL-style linking and various forms of API automation,
we need every route endpoint to have a unique name. The Operation models
a "<subject>.<verb>" pair as a name that can be mapped to a URL or href
at runtime.

### DIFF
--- a/microcosm_flask/conventions/health.py
+++ b/microcosm_flask/conventions/health.py
@@ -7,7 +7,6 @@ using HTTP 200/503 status codes to indicate healthiness.
 """
 from flask import jsonify
 
-from microcosm.api import defaults
 from microcosm_flask.naming import singleton_path_for
 from microcosm_flask.operations import Operation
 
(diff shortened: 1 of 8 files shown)


microcosm_flask/conventions/health.py
```diff
@@ -47,6 +51,9 @@ def to_dict(self):
         )
 
 
+@defaults(
+    path_prefix="/api",
```
jessemyers: I like the `/api` convention because it leaves room for serving `/admin` or static content.

microcosm_flask/conventions/health.py
```diff
@@ -56,8 +63,9 @@ def configure_health(graph):
     """
 
     health = Health(name=graph.metadata.name)
+    path = graph.config.health.path_prefix + singleton_path_for(Health)
 
-    @graph.flask.route("/api/health")
+    @graph.route(path, Operation.Retrieve, Health)
```
jessemyers: APIs that use the operation convention can use a slightly different route decorator that specifies the path, operation, and resource/class and does the rest according to conventions.

PR #6, opened 2016-03-10
Add support for linking, pagination, and resource discovery.

jessemyers:
Adds the convention that a service has an endpoint where you can find other top-level endpoints.

In the process, build out HAL JSON linking and some pagination support.


microcosm_flask/conventions/discovery.py
```diff
@@ -0,0 +1,60 @@
...
+def iter_search_operations(graph, operations):
+    for rule in graph.flask.url_map.iter_rules():
+        if "." in rule.endpoint:
+            operation, obj = Operation.parse(rule.endpoint)
+            if operation in operations:
+                yield operation, obj
+
+
+@defaults(
+    name="all",
```
jessemyers: The default behavior will be to serve a discovery endpoint at `/all` for all "search" endpoints. This can obviously be tuned. I agonized a bit over `/all` or just `/`

microcosm_flask/paging.py
```diff
@@ -0,0 +1,89 @@
...
+        Convert to tuples for deterministic order when passed to urlencode.
+
+        """
+        return [
+            ("offset", self.offset),
+            ("limit", self.limit),
+        ]
+
+
+class PaginatedList(object):
```
jessemyers: I'm not using the paginated list yet; it will get fleshed out as soon as I add the CRUD layer.

microcosm_flask/tests/conventions/test_discovery.py
```diff
@@ -0,0 +1,41 @@
...
+    is_,
+)
+
+from microcosm.api import create_object_graph
+from microcosm_flask.operations import Operation
+
+
+def test_discovery():
+    graph = create_object_graph(name="example", testing=True)
+    # TODO: can we make this automatic?
```
jessemyers: I ran into a design flaw here:
 - This code only references `route`, which only references the root `flask` app, not the fully configured `app` that includes `discovery`, `health`, etc.
 - I can't easily change `route` to reference `app` because it creates a circular dependency
 - If just using `route` is a common use pattern, we force users to call `graph.use()` on boilerplate that would otherwise be called when initialization `app`

Need to think about this more.

microcosm_flask/routing.py
```diff
@@ -29,7 +29,12 @@ def search_foo():
             pass
 
     """
+    graph.use(
```
jessemyers: Another graph awkwardness: if `uuid` isn't enforced here, the routes might fail (if they use uuids).

PR #7, opened 2016-03-10
CRUD convention

jessemyers:
Allow generation of CRUD APIs according to conventions that manage the basic encoding, decoding, and validation.

Underlying persistence is delegated to user-provided functions (which should line up well with the `Store` model in `microcosm-postgres`).


microcosm_flask/audit.py
```diff
@@ -39,6 +43,7 @@ def wrapper(*args, **kwargs):
             audit_dict.update(
                 success=False,
                 message=extract_error_message(error)[:2048],
+                context=extract_context(error),
```
jessemyers: We now have validation errors in our error response context.

microcosm_flask/tests/conventions/fixtures.py
```diff
@@ -0,0 +1,69 @@
...
+"""
+from uuid import uuid4
+
+from marshmallow import fields, Schema
+
+from microcosm_flask.linking import Links, Link
+from microcosm_flask.operations import Operation
+
+
+class Person(object):
```
jessemyers: This is a simple example of using the prerequisites for using the CRUD convention. It has:
 - a model (Person)
 - a schema for creating new people
 - a schema for a fully populated person
 - trivial/naive functions for each CRUD operation

PR #12, opened 2016-03-23
Support ad hoc command and query operations

jessemyers:
While microcosm aims for drive most operations from conventions, it has the plumbing to allow custom operations as well.

I've added two classes of operations and provided test cases that demonstrate how they (can) work:
 - Commands are verb-oriented, non-RESTful uses of POST
 - Queries are verb-oriented, non-RESTful uses of GET

### DIFF
--- a/microcosm_flask/swagger/definitions.py
+++ b/microcosm_flask/swagger/definitions.py
@@ -192,9 +192,15 @@ def add_responses(swagger_operation, operation, obj, func):
         description="An error occcurred",
         resource=type_name(name_for(ErrorSchema())),
     )
+
+    if hasattr(func, "__doc__"):
+        description = func.__doc__.strip().splitlines()[0]
+    else:
+        description = "{} {}".format(operation.value.name, name_for(obj))
+
     # resources response
     swagger_operation.responses[str(operation.value.default_code)] = build_response(
-        description=getattr(func, "__doc__") or "{} {}".format(operation.value.name, name_for(obj)),
+        description=description,
         resource=get_response_schema(func),
     )
 
--- a/microcosm_flask/tests/conventions/test_command.py
+++ b/microcosm_flask/tests/conventions/test_command.py
@@ -14,7 +14,6 @@ from microcosm.api import create_object_graph
 
 from microcosm_flask.conventions.encoding import dump_response_data, load_request_data
 from microcosm_flask.conventions.registry import request, response
-from microcosm_flask.conventions.swagger import register_swagger_endpoint
 from microcosm_flask.operations import Operation
 
 
@@ -51,7 +50,18 @@ def make_command(graph, request_schema, response_schema):
 class TestCommand(object):
 
     def setup(self):
-        self.graph = create_object_graph(name="example", testing=True)
+        # override configuration to use "query" operations for swagger
+        def loader(metadata):
+            return dict(
+                swagger_convention=dict(
+                    # default behavior appends this list to defaults; use a tuple to override
+                    operations=["command"],
+                    version="v1",
+                ),
+            )
+
+        self.graph = create_object_graph(name="example", testing=True, loader=loader)
+        self.graph.use("swagger_convention")
 
         make_command(self.graph, CommandArgumentSchema(), CommandResultSchema())
 
@@ -86,17 +96,6 @@ class TestCommand(object):
         Swagger definitions including this operation.
 
         """
-        base_path = "/api"
-        path_prefix = "/v1"
-
-        def match_func(operation, obj, rule):
-            return (
-                rule.rule.startswith(base_path + path_prefix) and
-                operation == Operation.Command
-            )
-
-        register_swagger_endpoint(self.graph, "swagger", "v1", path_prefix, match_func)
-
         response = self.client.get("/api/v1/swagger")
         assert_that(response.status_code, is_(equal_to(200)))
         swagger = loads(response.get_data())
--- a/microcosm_flask/tests/conventions/test_command.py
+++ b/microcosm_flask/tests/conventions/test_command.py
@@ -1,0 +1,131 @@
+"""
+Command operation tests.
+
+"""
+from json import dumps, loads
+from hamcrest import (
+    assert_that,
+    equal_to,
+    is_,
+)
+
+from marshmallow import fields, Schema
+from microcosm.api import create_object_graph
+
+from microcosm_flask.conventions.encoding import dump_response_data, load_request_data
+from microcosm_flask.conventions.registry import request, response
+from microcosm_flask.conventions.swagger import register_swagger_endpoint
+from microcosm_flask.operations import Operation
+
+
+class CommandArgumentSchema(Schema):
+    value = fields.String(required=True)
+
+
+class CommandResultSchema(Schema):
+    result = fields.Boolean(required=True)
+    value = fields.String(required=True)
+
+
+def make_command(graph, request_schema, response_schema):
+    """
+    Create an example command route.
+
+    """
+
+    @graph.route("/v1/foo/do", Operation.Command, "foo")
+    @request(request_schema)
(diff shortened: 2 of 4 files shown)


microcosm_flask/swagger/definitions.py
```diff
@@ -192,9 +192,15 @@ def add_responses(swagger_operation, operation, obj, func):
         description="An error occcurred",
         resource=type_name(name_for(ErrorSchema())),
     )
+
+    if hasattr(func, "__doc__"):
+        description = func.__doc__.strip().splitlines()[0]
```
jessemyers: The change here handles user-written doc strings that have newlines.

microcosm_flask/tests/conventions/test_command.py
```diff
@@ -0,0 +1,131 @@
...
+    value = fields.String(required=True)
+
+
+def make_command(graph, request_schema, response_schema):
+    """
+    Create an example command route.
+
+    """
+
+    @graph.route("/v1/foo/do", Operation.Command, "foo")
```
jessemyers: This is how a command might look:
 - the URI is user-defined, but need to have the right prefix if you want swagger to pick it up
 - the final "foo" argument is essentially a namespace; it should be different for every command
 - the `@request` and `@response` decorators tell the swagger layer what schemas are in use
 - the doc string gets picked up in the swagger definition
 - the function body uses boilerplate to decode the request and encode the response
 - a "real" command would do something interesting to generate the `response_data` from the `request_data`; this logic can be extended to use path variables; just look at the use of `merge_data()` in `conventions/crud.py`.

microcosm_flask/tests/conventions/test_command.py
```diff
@@ -0,0 +1,130 @@
...
+
+class TestCommand(object):
+
+    def setup(self):
+        # override configuration to use "query" operations for swagger
+        def loader(metadata):
+            return dict(
+                swagger_convention=dict(
+                    # default behavior appends this list to defaults; use a tuple to override
+                    operations=["command"],
```
jessemyers: We need to tell the swagger layer to consider "command" operations.

microcosm_flask/tests/conventions/test_command.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def test_swagger(self):
+        """
+        Swagger definitions including this operation.
+
+        """
+        response = self.client.get("/api/v1/swagger")
+        assert_that(response.status_code, is_(equal_to(200)))
+        swagger = loads(response.get_data())
+        assert_that(swagger["paths"], is_(equal_to({
```
jessemyers: The resulting swagger containers this section.

microcosm_flask/tests/conventions/test_query.py
```diff
@@ -0,0 +1,140 @@
...
+    """
+
+    @graph.route("/v1/foo/get", Operation.Query, "foo")
+    @qs(request_schema)
+    @response(response_schema)
+    def foo_query():
+        """
+        My doc string
+        """
+        request_data = load_query_string_data(request_schema)
```
jessemyers: Queries look a lot like commands except we pull data from the query string. There's an `@qs` decorator and a different request parsing function. We also build the schema for query strings from the default pagination schema.

microcosm_flask/tests/conventions/test_query.py
```diff
@@ -0,0 +1,140 @@
+"""
+Query operation tests.
```
jessemyers: Note that we'll often be able to extend the "search" operation for queries and should do that if possible. But if that really doesn't fit, we can do ad hoc queries.

adamhadani:
LGTM!

microcosm_flask/tests/conventions/test_command.py
srt32: What do you think about name spacing non-RESTful actions? The idea could be similar to https://github.com/interagent/http-api-design/blob/503b92f9db25f23f66bc512be86aac6f562c2427/en/requests/actions.md#L1

microcosm_flask/tests/conventions/test_command.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+class CommandResultSchema(Schema):
+    result = fields.Boolean(required=True)
+    value = fields.String(required=True)
+
+
+def make_command(graph, request_schema, response_schema):
+    """
+    Create an example command route.
```
srt32: On a more philosophical note, why make command endpoints as opposed to generating virtual RESTful resources?

 An example is https://globality.readme.io/docs/submit-scopeofwork, where the action is POST `.../submission` and it "creates" a "submission", which (as implementation details internal to the service) changes the state of the associated object and runs any needed side effects for the business.

microcosm_flask/tests/conventions/test_query.py
srt32: When would a developer reach for a custom query endpoint versus passing query params to the vanilla search (index) enpoint?

microcosm_flask/tests/conventions/test_command.py
```diff
@@ -0,0 +1,130 @@
...
+    def test_swagger(self):
+        """
+        Swagger definitions including this operation.
+
+        """
+        response = self.client.get("/api/v1/swagger")
+        assert_that(response.status_code, is_(equal_to(200)))
+        swagger = loads(response.get_data())
+        assert_that(swagger["paths"], is_(equal_to({
+            "/foo/do": {
```
srt32: Is it worth testing the whole swagger format here?

microcosm_flask/tests/conventions/test_command.py
jessemyers: I think it's a fine idea. Want to run with it?

There's a `naming.py` module that defines naming conventions for URI paths among other things. It would make sense to have a path generator for actions and queries and use that in these examples.

microcosm_flask/tests/conventions/test_command.py
jessemyers: Not really. It's tested elsewhere.

We could probably scale back this test to just validate that the right path/method/tags/operationId are defined; the rest should be boilerplate from the schema definitions. I didn't do that because I thought it'd be informative within this PR to demonstrate how the swagger comes out.

microcosm_flask/tests/conventions/test_query.py
jessemyers: I'd like to think that *neither* ad hoc command or query endpoints are needed and that we can always massage an API into something that is RESTful by inventing suitable resources.

I imagine queries would be used if the pain/overhead of adding a resource to search was too high.

microcosm_flask/tests/conventions/test_command.py
jessemyers: I'm with you. These endpoints are a nod to practicality, not "good" design. I also wouldn't use the term "virtual" here; I think that implies (incorrectly) that HTTP resources need to have a one-to-one mapping in the persistence layer.

This feature came about because we were discussing how to migrate some of the existing endpoints that are not RESTful. It's going to be less error prone to migrate them with their current paths than to reinterpret them in terms of new resources.

srt32: :ok_hand:

microcosm_flask/tests/conventions/test_query.py
srt32: Are you proposing a `Search` resource of some kind?  What's the difference between a list endpoint that accepts query params and a search endpoint? Would the search endpoint allow for more custom functionality of some kind?

jessemyers: The Search "operation" already exists and is part of the CRUD convention. It's the name I gave to a `GET` against the collection of resources and it can have query parameters -- by default it supports paging parameters, but this can be extended.

I think this is the same as a "list endpoint".

PR #15, opened 2016-03-30
Add support for a 'relational' create operation

adamhadani:
TODO:
- Add unit-tests for the whole relational conventions which didn't exist previously

### DIFF
--- a/microcosm_flask/tests/conventions/test_crud.py
+++ b/microcosm_flask/tests/conventions/test_crud.py
@@ -101,6 +101,29 @@ class TestCrud(object):
             },
         })
 
+    def test_create_empty_object(self):
+        response = self.client.post("/api/person", data='{}')
+        self.assert_response(response, 422, {
+            "code": 422,
+            "message": "Validation error",
+            "retryable": False,
+            "context": {
+                "errors": [{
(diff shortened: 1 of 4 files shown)


microcosm_flask/tests/conventions/test_crud.py
```diff
@@ -101,6 +101,29 @@ def test_create(self):
             },
         })
 
+    def test_create_empty_object(self):
+        response = self.client.post("/api/person", data='{}')
```
adamhadani: This does not test exactly the use case we saw was breaking - however I could not for the life of me figure out how to reproduce a successful POST with 'empty' body similar to what httpie is doing..

microcosm_flask/tests/conventions/test_crud.py
```diff
@@ -101,6 +101,29 @@ def test_create(self):
...
+        self.assert_response(response, 422, {
+            "code": 422,
+            "message": "Validation error",
+            "retryable": False,
+            "context": {
+                "errors": [{
```
jessemyers: Maybe use {{contains_inanyorder}} on these errors so that ordering changes doesn't break us?

jessemyers:
LGTM if you address test fragility

PR #21, opened 2016-04-01
Ensure that `sync-resources` looks at embedded links.

jessemyers:
Without this change, I was missing some links when starting from certain URIs.


adamhadani:
LGTM

PR #22, opened 2016-04-03
Introduce a namespace abstraction to simplify endpoint registration.

jessemyers:
Earlier versions of this library put a lot of responsiblity into the Operation enum
to define naming conventions. Since the naming depends heavily on the context of the
resources being operated on, the Operation functions had some fairly complex argumentation.

In the new approach, these functions move to a Namespace class, which has concrete
instance variables to make the naming evaluation simpler. As much as possible, convention
creation functions now accept Namespace instances, though not yet everywhere because of
backward compatibility concerns.

### DIFF
--- a/microcosm_flask/conventions/crud.py
+++ b/microcosm_flask/conventions/crud.py
@@ -13,7 +13,7 @@ from microcosm_flask.conventions.encoding import (
     require_response_data,
 )
 from microcosm_flask.conventions.registry import qs, request, response
-from microcosm_flask.naming import collection_path_for, instance_path_for, name_for
+from microcosm_flask.namespaces import Namespace
 from microcosm_flask.operations import Operation
 from microcosm_flask.paging import Page, PaginatedList, make_paginated_list_schema
 
@@ -34,13 +34,12 @@ def _crud(operation):
 
 
 @_crud(Operation.Search)
-def register_search_endpoint(graph, obj, path_prefix, func, request_schema, response_schema):
+def register_search_endpoint(graph, ns, func, request_schema, response_schema):
     """
     Register a search endpoint.
 
     :param graph: the object graph
-    :param obj: the target resource or resource name
-    :param path_prefix: the routing path prefix
+    :param ns: the namespace
     :param func: a search function, which must:
       - accept kwargs for the query string (minimally for pagination)
       - return a tuple of (items, count) where count is the total number of items
@@ -49,9 +48,9 @@ def register_search_endpoint(graph, obj, path_prefix, func, request_schema, resp
     :param response_schema: a marshmallow schema to encode (a single) response item
     """
 
-    paginated_list_schema = make_paginated_list_schema(obj, response_schema)()
+    paginated_list_schema = make_paginated_list_schema(ns, response_schema)()
 
-    @graph.route(path_prefix + collection_path_for(obj), Operation.Search, obj)
+    @graph.route(ns.collection_path, Operation.Search, ns)
     @qs(request_schema)
     @response(paginated_list_schema)
     def search(**path_data):
@@ -60,20 +59,19 @@ def register_search_endpoint(graph, obj, path_prefix, func, request_schema, resp
         items, count = func(**merge_data(path_data, request_data))
         # TODO: use the schema for encoding
         return jsonify(
-            PaginatedList(obj, page, items, count, response_schema).to_dict()
+            PaginatedList(ns, page, items, count, response_schema).to_dict()
         )
 
-    search.__doc__ = "Search the collection of all {}".format(pluralize(name_for(obj)))
+    search.__doc__ = "Search the collection of all {}".format(pluralize(ns.subject_name))
 
 
 @_crud(Operation.Create)
-def register_create_endpoint(graph, obj, path_prefix, func, request_schema, response_schema):
+def register_create_endpoint(graph, ns, func, request_schema, response_schema):
     """
     Register a create endpoint.
 
     :param graph: the object graph
-    :param obj: the target resource or resource name
-    :param path_prefix: the routing path prefix
+    :param ns: the namespace
     :param func: a create function, which must:
       - accept kwargs for the request and path data
       - return a new item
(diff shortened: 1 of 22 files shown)


microcosm_flask/conventions/crud.py
```diff
@@ -34,13 +34,12 @@ def decorator(func):
 
 
 @_crud(Operation.Search)
-def register_search_endpoint(graph, obj, path_prefix, func, request_schema, response_schema):
+def register_search_endpoint(graph, ns, func, request_schema, response_schema):
```
jessemyers: All of the internal route registration functions now take a `Namespace`.

This simplifies the call arguments a bit.

microcosm_flask/conventions/crud.py
```diff
@@ -209,6 +204,7 @@ def configure_crud(graph, obj, mappings, path_prefix=""):
         }
 
     """
+    ns = Namespace.make(ns, path=path_prefix)
```
jessemyers: The older `configure_crud` invocations would take strings, classes, etc. We now expect a `Namespace`, but still support the older calls for compatibility via some logic in `Namespace.make()`

microcosm_flask/conventions/crud.py
```diff
@@ -49,9 +48,9 @@ def register_search_endpoint(graph, obj, path_prefix, func, request_schema, resp
     :param response_schema: a marshmallow schema to encode (a single) response item
     """
 
-    paginated_list_schema = make_paginated_list_schema(obj, response_schema)()
+    paginated_list_schema = make_paginated_list_schema(ns, response_schema)()
 
-    @graph.route(path_prefix + collection_path_for(obj), Operation.Search, obj)
+    @graph.route(ns.collection_path, Operation.Search, ns)
```
jessemyers: All calls to the internal router now expect a `path`, an `Operation`, and ` Namespace`. In most cases, the `path` should be obtained from a `Namespace` property.

microcosm_flask/conventions/relation.py
```diff
@@ -77,9 +76,9 @@ def register_search_relation_endpoint(graph, obj, path_prefix, func, request_sch
     :param response_schema: a marshmallow schema to encode (a single) response item
     """
 
-    paginated_list_schema = make_paginated_list_schema(obj[1], response_schema)()
+    paginated_list_schema = make_paginated_list_schema(ns.object_ns, response_schema)()
```
jessemyers: Note the `ns.object_ns` argument. Here's what's happening:
 - The pagination and linking code now depend on `Namespaces`, so they expect us to pass one
 - For relation-oriented endpoints, we have a subject, verb, object construction (e.g. `folder.search_for_files`)
 - The (paginated) result type of `search_for` operation is the object, not the subject (e.g. returns a page of `files`, not `folders`)
 - The `Namespace` knows how to perform this transformation

microcosm_flask/conventions/relation.py
```diff
@@ -99,6 +98,12 @@ def configure_relation(graph, from_obj, to_obj, mappings, path_prefix=""):
     Register relation endpoint(s) between two resources.
 
     """
+    ns = Namespace(
```
jessemyers: Unlike the CRUD configuration function, there's no good way to have the caller pass a `Namespace` without changing the call signature in a backwards incompatible way.

My guess is that we'll choose to break compatibly (it's early days for this library), but I didn't want to do so without discussion.

microcosm_flask/namespaces.py
```diff
@@ -0,0 +1,154 @@
...
+
+    @property
+    def relation_path(self):
+        return self.path + relation_path_for(self.subject, self.object_)
+
+    @property
+    def singleton_path(self):
+        return self.path + singleton_path_for(self.subject)
+
+    def endpoint_for(self, operation):
```
jessemyers: One of the motivations of the `Namespace` abstraction is that it gives us a good place to inject alternate logic for endpoint naming.

Example: we've talked a bit about having low-level and high-level APIs (for the same resource). If we had such a thing for a delete operation, we'd currently end up with **two** endpoint names called `foo.delete`, which Flask will not allow. The path forward is to give the `Namespace` another member variable and generate different endpoint names based on some prefix (e.g. `x.foo.delete` and `y.foo.delete`).

microcosm_flask/namespaces.py
```diff
@@ -0,0 +1,154 @@
...
+        else:
+            return operation.value.pattern.format(
+                self.subject_name,
+                operation.value.name,
+            )
+
+    @staticmethod
+    def parse_endpoint(name):
+        """
+        Convert an endpoint name into an (operation, ns) tuple.
```
jessemyers: This is the inverse of the previous function; we need it when implementing the discovery or swagger convention, which iterate through (a subset of) all matching endpoints and use the underlying metadata to advertise APIs.

microcosm_flask/namespaces.py
```diff
@@ -0,0 +1,154 @@
...
+
+        """
+        parts = name.split(".")
+        operation = Operation.from_name(parts[1])
+        if len(parts) > 2:
+            return operation, Namespace(subject=parts[0], object_=parts[2])
+        else:
+            return operation, Namespace(subject=parts[0])
+
+    def url_for(self, operation, **kwargs):
```
jessemyers: The `url_for` and `href_for` functions were formerly part of the `Operation` enum.

microcosm_flask/swagger/naming.py
```diff
@@ -21,11 +19,11 @@ def operation_name(operation, obj):
         foo.search_for.bar => client.foo.search_for_bars()
 
     """
-    if isinstance(obj, (list, tuple)):
-        verb, rest = operation.value.name, obj[1:]
+    verb = operation.value.name
+    if ns.object_:
```
jessemyers: This is one of the few places that the internals of `Namespace` leak out. I wasn't ready to make the `Namespace` responsible for knowledge of Swagger naming conventions.

adamhadani:
LGTM - is there a full usage example somewhere to demonstrate it in Action / how one would write using it? Eg we'll update the cookie cutter project ?

jessemyers:
Here's a [PR of the cookiecutter changes](https://github.com/globality-corp/cookiecutter-microcosm-service/pull/8). The unit tests for the various conventions have also been updated here.

jessemyers:
Attention: @srt32 @adamhadani 

I added one more commit which creates a *breaking* change on the relation convention; it now requires either that a `Namespace` be passed in or that the subject/object be passed as a tuple. Without this commit, there isn't a great way to use namespaces for relation APIs.

PR #23, opened 2016-04-07
Add a timestamp field.

jessemyers:
Decodes isoformat and numeric timestamps; encodes one or the other based on config.

### DIFF
--- a/microcosm_flask/fields.py
+++ b/microcosm_flask/fields.py
@@ -2,7 +2,10 @@
 Custom fields.
 
 """
-from marshmallow.fields import Field
+from datetime import datetime
+
+from dateutil import parser
+from marshmallow.fields import Field, ValidationError
 
 
 class EnumField(Field):
@@ -46,3 +49,48 @@ class EnumField(Field):
             return getattr(self.enum, value)
         except AttributeError:
             self.fail('by_name', name=value)
+
+
+class TimestampField(Field):
+    """
+    Timestamp valued field, as either a unix timestamp (default) or isoformat string.
+
+    """
+    EPOCH = datetime(1970, 1, 1)
+
+    def __init__(self, use_isoformat=False, *args, **kwargs):
+        self.use_isoformat = use_isoformat
+        super(TimestampField, self).__init__(*args, **kwargs)
+
+    def _serialize(self, value, attr, obj):
+        """
+        Serialize value as a timestamp, either as a Unix timestamp (in float second) or a UTC isoformat string.
+
+        """
+        if value is None:
+            return None
+
+        if self.use_isoformat:
+            return datetime.utcfromtimestamp(value).isoformat()
+        else:
+            return value
+
+    def _deserialize(self, value, attr, obj):
+        """
+        Deserialize value as a Unix timestamp (in float seconds).
+
+        Handle both numeric and UTC isoformat strings.
+
+        """
+        if value is None:
+            return None
+
+        try:
+            return float(value)
+        except ValueError:
+            parsed = parser.parse(value)
+            if parsed.tzinfo:
+                if parsed.utcoffset().total_seconds():
+                    raise ValidationError("Timestamps must be defined in UTC")
+                parsed = parsed.replace(tzinfo=None)
+            return (parsed - TimestampField.EPOCH).total_seconds()
--- a/microcosm_flask/tests/test_fields.py
+++ b/microcosm_flask/tests/test_fields.py
@@ -1,0 +1,131 @@
+"""
+Test fields.
+
+"""
+from hamcrest import (
+    assert_that,
+    contains,
+    equal_to,
+    is_,
+)
+from marshmallow import Schema
+
+from microcosm_flask.fields import TimestampField
+
+
+TIMESTAMP = 1427702400.0
+TIMESTAMP_EXTRA_PRECISION = 1427702400.1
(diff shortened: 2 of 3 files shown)


microcosm_flask/fields.py
```diff
@@ -46,3 +49,48 @@ def _deserialize_by_name(self, value, attr, data):
...
+
+    def _serialize(self, value, attr, obj):
+        """
+        Serialize value as a timestamp, either as a Unix timestamp (in float second) or a UTC isoformat string.
+
+        """
+        if value is None:
+            return None
+
+        if self.use_isoformat:
```
jessemyers: The assumption here is that the internal value is always a Unix timestamp (in float seconds). We either return it verbatim (as a float) or encode it as an iso string.

It would be possible to generalize this to support datetimes internally, but I'd rather set the precedent that internal timestamps are numbers.

microcosm_flask/fields.py
```diff
@@ -46,3 +49,48 @@ def _deserialize_by_name(self, value, attr, data):
...
+        """
+        if value is None:
+            return None
+
+        try:
+            return float(value)
+        except ValueError:
+            parsed = parser.parse(value)
+            if parsed.tzinfo:
+                if parsed.utcoffset().total_seconds():
```
jessemyers: It's possible to parse non-UTC timestamps, but it's probably a sign that something upstream isn't in UTC, so I chose to raise an error.

microcosm_flask/tests/test_fields.py
```diff
@@ -0,0 +1,131 @@
...
+
+    """
+    schema = TimestampSchema()
+    result = schema.dump({
+        "unix": TIMESTAMP,
+        "iso": TIMESTAMP,
+    })
+
+    assert_that(result.data["unix"], is_(equal_to(TIMESTAMP)))
+    assert_that(result.data["iso"], is_(equal_to(ISOFORMAT_NAIVE)))
```
jessemyers: It's worth noting that this output doesn't have the "Z" at the end. I couldn't find a good way to enforce that (but didn't try that hard).

microcosm_flask/fields.py
srt32: Why does `if parsed.utcoffset().total_seconds():` signify a parsing error?

setup.py
```diff
@@ -25,6 +25,8 @@
         "microcosm>=0.5.1",
         "microcosm-logging>=0.2.0",
         "openapi>=0.2.0",
+        "python-dateutil>=2.5.2",
+        "pytz>=2016.3",
```
srt32: This dep gets used behind the scenes?

microcosm_flask/fields.py
jessemyers: Right. Iso format allows "+/-HH:MM" instead of "Z".

I could parse and compute the offset, but I'm considering it a failure if we have inbound messages that are not UTC.

setup.py
jessemyers: `pytz` doesn't; I added it because I was trying to force utc. I'll remove it.

microcosm_flask/fields.py
```diff
@@ -46,3 +49,48 @@ def _deserialize_by_name(self, value, attr, data):
...
+    def _serialize(self, value, attr, obj):
+        """
+        Serialize value as a timestamp, either as a Unix timestamp (in float second) or a UTC isoformat string.
+
+        """
+        if value is None:
+            return None
+
+        if self.use_isoformat:
+            return datetime.utcfromtimestamp(value).isoformat()
```
srt32: @Phylodome is getting some `year is out of range for ` errors on jeremiah right now and I think it uses similar code.  Probably worth our figuring out what that issue is prior to merging.

jessemyers: Do you have a sample input? Unit tests are good for this sort of thing.

microcosm_flask/fields.py
```diff
@@ -46,3 +49,48 @@ def _deserialize_by_name(self, value, attr, data):
...
+        Deserialize value as a Unix timestamp (in float seconds).
+
+        Handle both numeric and UTC isoformat strings.
+
+        """
+        if value is None:
+            return None
+
+        try:
+            return float(value)
```
srt32: This case is handling if we were passed a unix timestamp already?

jessemyers: Yup. I'd like to preserve the ability to avoid isoformat entirely.

srt32:
LGTM

microcosm_flask/fields.py
srt32: What's the diff between unix timestamps that are ints and unix timestamps that have a decimal (float)? I ran into a couple surprising results when I was experimenting earlier

allisonfong:
This is great! Thanks!

microcosm_flask/fields.py
jessemyers: Historically, C-based systems defined timestamps as seconds since the Jan 1 1970 UTC. People needed a way to extend this sub-second timing. The C world added a struct with both seconds and nanoseconds. Java switched to a long in milliseconds. Python opted for floating point seconds.

Since this is Python, I'm using the floating point convention. I don't think there's a universal answer, but float is nice because it's trivial to case to seconds for systems that need that and is easily encoded as a single value, whether in a database or over the network as JSON.

PR #24, opened 2016-04-12
Refactor conventions to use a common model (base class + data object)

jessemyers:
Conventions are recipes for mapping some user data ("EndpointDefinitions") to operations attached at a specific namespace.

This is now explicit.

As a user, you do something like:

```
convention = CRUDConvention(graph)

mapping = {
    Operation.Create: (func, request_schema, response_schema),
}

convention.configure(ns, mapping)
```

For readability, it's also now possible to use a clearer structure for the mapping:

```
mapping = dict(
    create=EndpointDefinition.make(func, request_schema, response_schema),
)
```

Overall, this cleanup will make it easier to define behaviors that apply to all conventions. In particular, I want to extend the `Namespace` to support the idea of a low and high level path prefix and allow the same (or very similar) logic to be applied to each of these.

### DIFF
--- a/microcosm_flask/conventions/base.py
+++ b/microcosm_flask/conventions/base.py
@@ -1,0 +1,99 @@
+"""
+Convention base class.
+
+"""
+from microcosm_flask.operations import Operation
+
+
+class EndpointDefinition(tuple):
+    """
+    A definition for an endpoint.
+
+    """
+    @classmethod
+    def make(cls, func=None, request_schema=None, response_schema=None):
+        """
+        :param func: a function to process request data and return response data
+        :param request_schema: a marshmallow schema to decode/validate request data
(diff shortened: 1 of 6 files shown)


microcosm_flask/conventions/base.py
```diff
@@ -0,0 +1,99 @@
...
+
+        The input might be an `Operation` enum or a string.
+
+        """
+        if isinstance(operation, Operation):
+            operation_name = operation.name.lower()
+        else:
+            operation_name = operation.lower()
+
+        return getattr(self, "configure_{}".format(operation_name))
```
jessemyers: Conventions indicate support for a specific operation by implementing the `configure_foo` member function (where "foo" is the operation name)

adamhadani:
LGTM

PR #25, opened 2016-04-13
Add tests for enum (de)serialization.

jessemyers:
Ensure that stringified int enum values work because that's what happens when you pass integers via a query string...


cdm-ium:
+1

PR #26, opened 2016-04-14
Overload tuple's __new__ instead of defining custom factory method

jessemyers:
Small cleanup to have more idiomatic code


adamhadani:
LGTM

PR #27, opened 2016-04-14
Centralize path management into Namespace and routing layers.

jessemyers:
This is a) good cleanup and b) paves the way for a namespace having multiple paths for low/high level APIs.

### DIFF
--- a/microcosm_flask/conventions/swagger.py
+++ b/microcosm_flask/conventions/swagger.py
@@ -11,14 +11,11 @@ from microcosm_flask.conventions.base import Convention
 from microcosm_flask.conventions.registry import iter_endpoints
 from microcosm_flask.namespaces import Namespace
 from microcosm_flask.operations import Operation
+from microcosm_flask.routing import make_path
 from microcosm_flask.swagger.definitions import build_swagger
 
 
 class SwaggerConvention(Convention):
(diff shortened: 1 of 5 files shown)


microcosm_flask/conventions/swagger.py
```diff
@@ -50,7 +48,7 @@ def configure_discover(self, ns, definition):
         """
         @self.graph.route(ns.singleton_path, Operation.Discover, ns)
         def discover():
-            swagger = build_swagger(self.graph, ns.version, ns.path, self.operations)
```
jessemyers: First main change: the swagger layer had its own path calculation; it now delegates to the `Namespace` or the `make_path()` function from the routing layer

microcosm_flask/namespaces.py
```diff
@@ -31,13 +31,26 @@ class Namespace(object):
...
+        :param version: the version of this namespace
+        """
         self.subject = subject
         self.object_ = object_
-        self.path = path
+        self._path = path or ""
         self.version = version
 
     @property
+    def path(self):
```
jessemyers: Second main change: the `Namespace` has a compute path attribute

adamhadani:
LGTM

PR #28, opened 2016-04-18
Fix up path handling for swagger with new convention and namespace logic

jessemyers:
The previous code used the wrong namespace to determine whether an endpoint should be advertised via swagger. It also was consistent in how paths were computed.

### DIFF
--- a/microcosm_flask/conventions/swagger.py
+++ b/microcosm_flask/conventions/swagger.py
@@ -24,8 +24,7 @@ class SwaggerConvention(Convention):
             for operation_name in self.graph.config.swagger_convention.operations
         }
 
-    @property
-    def operations(self):
+    def find_matching_endpoints(self, swagger_ns):
         """
         Compute current matching endpoints.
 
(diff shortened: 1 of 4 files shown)


microcosm_flask/conventions/swagger.py
```diff
@@ -35,7 +34,7 @@ def operations(self):
         def match_func(operation, ns, rule):
             # only expose endpoints that have the correct path prefix and operation
             return (
-                rule.rule.startswith(make_path(self.graph, ns.path)) and
+                rule.rule.startswith(make_path(self.graph, swagger_ns.path)) and
```
jessemyers: This was a bug; during the convention refactor, I accidentally switched which namespace was being used to compute whether an endpoint should be advertised with swagger.

microcosm_flask/swagger/definitions.py
```diff
@@ -36,7 +37,7 @@ def build_swagger(graph, ns, operations):
     Build out the top-level swagger definition.
 
     """
-    base_path = graph.config.route.path_prefix
+    base_path = make_path(graph, ns.path)
```
jessemyers: This just makes path construction simpler and more consistent.

microcosm_flask/tests/swagger/test_definitions.py
```diff
@@ -30,7 +30,6 @@
 def test_build_swagger():
     graph = create_object_graph(name="example", testing=True)
     ns = Namespace(
-        path="/v1",
```
jessemyers: This was another bug; in the namespace logic, the version input is used to build the actual endpoint path.

(This is still confusing and deserves another round of cleanup and some unit tests.)

PR #29, opened 2016-04-21
Add a RetrieveFor operation: this is the equivalent for a SearchFor for a singleton

jessemyers:


amitbn:
LGTM
Just want to make sure I understand the purpose - 
That's exactly as same as SearchFor, but the returning list of `items` would be always of size of 1?

amitbn:
+1

amitbn:
+1

PR #30, opened 2016-04-22
Add support for Swagger mapping from marshmallow Fields.Dict() type

adamhadani:
For APIs which work on top of non-relational / document-oriented persistence (e.g DynamoDB), there are cases where a fairly opaque payload is involved in CRUD operations (e.g a JSON / dict object mapped to a field with no strict schema).
Marshmallow already supports this via the fields.Dict type, and Swagger supports a generic 'object' field type definition.

This PR adds this mapping to microcosm-flask explicitly


jessemyers:
+1

PR #31, opened 2016-04-22
Allow exceptions to trigger non-errors

jessemyers:
I'm using exceptions to abort workflows that are not errors, e.g. returning a 202 for an in-progress response.

This change ensures that audit logs don't indicate errors or stack traces in this case


adamhadani:
LGTM

PR #32, opened 2016-04-25
add automatic testing in CI on py3.5

adamhadani:


jessemyers:
I think the build errors can be solved using `six`

PR #34, opened 2016-05-02
GLOB-2272-add-request-id-to-request-logging

sethisernhagen:
https://globality.atlassian.net/browse/GLOB-2272

- add X-Request-Id to audit log

### DIFF
--- a/microcosm_flask/audit.py
+++ b/microcosm_flask/audit.py
@@ -28,6 +28,7 @@ def audit(func):
     @wraps(func)
     def wrapper(*args, **kwargs):
         response = None
+        request_id = request.headers.get('X-Request-Id', 'None')
 
         # always include these fields
         audit_dict = dict(
@@ -35,6 +36,7 @@ def audit(func):
             func=func.__name__,
             method=request.method,
         )
+        audit_dict.update({"X-Request-Id": request_id})
 
         # include request body on debug (if any)
         if current_app.debug and request.get_json(force=True, silent=True):


microcosm_flask/audit.py
```diff
@@ -28,13 +28,15 @@ def audit(func):
     @wraps(func)
     def wrapper(*args, **kwargs):
         response = None
+        request_id = request.headers.get('X-Request-Id', 'None')
```
jessemyers: Because `microcosm` is open source, it's a good idea to be aware of how our assumptions might appear to third parties using this software. In this case, there are two oddities:

1. We are logging this particular header and no others
2. We are using a default value of the *string* 'None' (vs `None`) or just omitting the header entirely.

A better design would make the list of headers to log configurable somehow and only log the headers that are present.

The second part is easier:

```
logged_headers = {
    key: value
    for key, value in request.headers.items()
    if should_log_header(key)
}
...

audit_dict.update(logged_headers)
```

The harder part is specifying the `should_log_header` function. We want this to be configurable, so you'll want to extend the binding function (`configure_audit_decorator`, below) to take this configuration. We also need to pass the configuration to the `audit` decorator, which isn't easy unless we wrap the `audit` decorator function in another scope that is aware of this function.

So:

```
def create_audit_decorator(should_log_header):
    # same audit function, indented here
    def audit(func):
        ...
    return audit
```

And:

```
...

@defaults(
    header_prefix="X-Request",
)
def configure_audit_decorator(graph):
    def should_log_headers(key):
         return key.startswith(graph.config.audit.header_prefix)

    return create_audit_decorator(should_log_headers)
```

jessemyers: FYI, we now have code depending on the `audit` function. We can either fix whatever breaks when we change this or do something a little less clean, such as using `setattr` to pass the `should_log_headers` func to `audit`.

PR #35, opened 2016-05-20
Implement audit logging options.

jessemyers:
This PR allows audit logging to be configurable while retaining backwards compatibility. As part of making audit logging configurable, it now supports logging headers based on configuration.

This supersedes #34

### DIFF
--- a/microcosm_flask/audit.py
+++ b/microcosm_flask/audit.py
@@ -2,18 +2,38 @@
 Audit log support for Flask routes.
 
 """
-from logging import getLogger
+from collections import namedtuple
 from functools import wraps
+from logging import getLogger
 from json import loads
 from traceback import format_exc
 
 from flask import current_app, g, request
-
+from microcosm.api import defaults
 from microcosm_flask.errors import (
     extract_context,
     extract_error_message,
     extract_status_code,
 )
+
+
+X_REQUEST = "X-Request"
+
+AuditOptions = namedtuple("AuditOptions", [
+    "include_request_body",
+    "include_response_body",
+    "include_header",
+])
+
+
+def include_header_matching_prefix(prefix):
+    """
+    Match all headers starting with prefix.
+
+    """
+    def include_header(header):
+        return header.startswith(prefix)
+    return include_header
 
 
 def audit(func):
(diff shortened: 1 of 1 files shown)


microcosm_flask/audit.py
```diff
@@ -2,80 +2,131 @@
...
 from microcosm_flask.errors import (
     extract_context,
     extract_error_message,
     extract_status_code,
 )
 
 
+X_REQUEST = "X-Request"
+
+AuditOptions = namedtuple("AuditOptions", [
```
jessemyers: There are three options supported now.

microcosm_flask/audit.py
```diff
@@ -2,80 +2,131 @@
...
+    """
+    Match all headers starting with prefix.
+
+    """
+    def include_header(header):
+        return header.startswith(prefix)
+    return include_header
+
+
 def audit(func):
```
jessemyers: I've retained the original `audit` function because some projects that have not fully ported to microcosm import this directly.

microcosm_flask/audit.py
```diff
@@ -2,80 +2,131 @@
...
-            )
 
-            # include response body on debug (if any)
-            if current_app.debug and body and not g.get("hide_body"):
-                try:
-                    audit_dict["response_body"] = loads(body)
-                except (TypeError, ValueError):
-                    # not json
-                    audit_dict["response_body"] = body
+def _audit_request(options, func, *args, **kwargs):
```
jessemyers: The body of the old audit function has been refactored to here; it takes options.

microcosm_flask/audit.py
```diff
@@ -2,80 +2,131 @@
...
+
+    # always include these fields
+    audit_dict = dict(
+        operation=request.endpoint,
+        func=func.__name__,
+        method=request.method,
+    )
+
+    # include request body on debug (if any)
+    if all((
```
jessemyers: We use the options here to control logging the request. There was also a bug here where we didn't respect `g.hide_body` for per-request auditing.

microcosm_flask/audit.py
```diff
@@ -2,80 +2,131 @@
...
+        current_app.debug,
+        options.include_request_body,
+        request.get_json(force=True, silent=True),
+        not g.get("hide_body"),
+    )):
+        audit_dict["request_body"] = request.get_json(force=True)
+
+    # include headers (conditionally)
```
jessemyers: New feature: logging matching headers

microcosm_flask/audit.py
```diff
@@ -2,80 +2,131 @@
...
+
+        audit_dict.update(
+            success=True,
+            status_code=status_code,
+        )
+
+        # include response body on debug (if any)
+        if all((
+                current_app.debug,
+                options.include_response_body,
```
jessemyers: We use the options again here.

microcosm_flask/audit.py
```diff
@@ -95,6 +146,11 @@ def parse_response(response):
         return response, 200
 
 
+@defaults(
```
jessemyers: The factory now takes defaults and constructs its own copy of the decorator that passes these in as options.

microcosm_flask/audit.py
```diff
@@ -2,80 +2,131 @@
...
+    )
+
+    # include request body on debug (if any)
+    if all((
+        current_app.debug,
+        options.include_request_body,
+        request.get_json(force=True, silent=True),
+        not g.get("hide_body"),
+    )):
+        audit_dict["request_body"] = request.get_json(force=True)
```
sethisernhagen: Why convert this to a dict?  Isn't this the cause for invalid json being logged? ie. u'

sethisernhagen:
+1 Looks great. Thanks Jesse!

microcosm_flask/audit.py
jessemyers: I'm not sure. Do you have a test case?

PR #36, opened 2016-05-31
Fixes for Flask 0.11 compatibility

jessemyers:
- Stop using deprecated `flask.ext.foo`
 - Use functions to register error handlers instead of modifying the underlying datastructure
   (which has changed semantics)

### DIFF
--- a/microcosm_flask/errors.py
+++ b/microcosm_flask/errors.py
@@ -127,10 +127,9 @@ def configure_error_handlers(graph):
     Register error handlers.
 
     """
-
     # override all of the werkzeug HTTPExceptions
     for code in default_exceptions.keys():
-        graph.flask.error_handler_spec[None][code] = make_json_error
+        graph.flask.register_error_handler(code, make_json_error)
 
     # register catch all for user exceptions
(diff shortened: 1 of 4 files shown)


microcosm_flask/errors.py
```diff
@@ -127,10 +127,9 @@ def configure_error_handlers(graph):
     Register error handlers.
 
     """
-
     # override all of the werkzeug HTTPExceptions
     for code in default_exceptions.keys():
-        graph.flask.error_handler_spec[None][code] = make_json_error
+        graph.flask.register_error_handler(code, make_json_error)
```
srt32: This API seems much nicer ;)

srt32:
LGTM

setup.py
```diff
@@ -17,7 +17,7 @@
     keywords="microcosm",
     install_requires=[
         "enum34>=1.1.2",
-        "Flask>=0.10.1",
+        "Flask>=0.11",
```
jessemyers: I'm not totally sure this is a good idea. I'm also not totally sure that the error handling registration functions worked the same way in `0.10.1`. A cursory read of the source only convinced me that they are *different*.

PR #37, opened 2016-06-03
Introduce an UpdateBatch convention

jessemyers:
Uses HTTP PATCH on a resource collection to apply several changes at once.

Most of heavy lifting is left to implementing services which will need to:
 -  Define suitable request and response schemas
 -  Determine how best to handle create-vs-update semantics
 -  Determine how best to report partial errors

### DIFF
--- a/microcosm_flask/conventions/crud_adapter.py
+++ b/microcosm_flask/conventions/crud_adapter.py
@@ -46,3 +46,22 @@ class CRUDStoreAdapter(object):
         identifier = kwargs.pop(self.identifier_key)
         model = self.store.model_class(id=identifier, **kwargs)
         return self.store.update(identifier, model)
+
+    def update_batch(self, **kwargs):
+        """
+        Simplistic batch update operation implemented in terms of `replace()`.
+
+        Assumes that:
+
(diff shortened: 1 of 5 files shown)


microcosm_flask/conventions/crud_adapter.py
```diff
@@ -46,3 +46,22 @@ def update(self, **kwargs):
         identifier = kwargs.pop(self.identifier_key)
         model = self.store.model_class(id=identifier, **kwargs)
         return self.store.update(identifier, model)
+
+    def update_batch(self, **kwargs):
```
jessemyers: The CRUD adapter's update batch function is probably too simplistic to be useful; we'll have to learn from experience.

microcosm_flask/tests/conventions/fixtures.py
```diff
@@ -45,6 +53,15 @@ def person_search(offset, limit):
     return [PERSON_1], 1
 
 
+def person_update_batch(items):
```
jessemyers: Also very simplistic.

amitbn:
+1

PR #38, opened 2016-06-06
add DeleteFor and ReplaceFor relation operations; add require_response_data decorator to relational CRUD operators

adamhadani:
- adds the two missing operations for the 'relation' Operator support: Delete and Replace.
- adds the require_response_data decorator that's been used in the main CRUD Operator commands to the relational equivalent

### DIFF
--- a/microcosm_flask/conventions/relation.py
+++ b/microcosm_flask/conventions/relation.py
@@ -75,6 +75,12 @@ class RelationConvention(Convention):
         """
         Register a replace-for relation endpoint.
 
+        For typical usage, this relation is not strictly required; once an object exists and has its own ID,
+        it is better to operate on it directly via dedicated CRUD routes.
+        However, in some cases, the composite key of (subject_id, object_id) is required to look up the object.
+        This happens, for example, when using DynamoDB where an object which maintains both a hash key and a range key
+        requires specifying them both for access.
+
         The definition's func should be a replace function, which must:
         - accept kwargs for the new-or-updated instance parameters
         - return the created or updated instance
--- a/microcosm_flask/conventions/relation.py
+++ b/microcosm_flask/conventions/relation.py
@@ -15,6 +15,7 @@ from microcosm_flask.conventions.encoding import (
     load_query_string_data,
     load_request_data,
     merge_data,
+    require_response_data,
 )
 from microcosm_flask.conventions.registry import qs, request, response
 from microcosm_flask.namespaces import Namespace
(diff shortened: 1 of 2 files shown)


microcosm_flask/conventions/relation.py
```diff
@@ -41,11 +42,56 @@ def configure_createfor(self, ns, definition):
         @response(definition.response_schema)
         def create(**path_data):
             request_data = load_request_data(definition.request_schema)
-            response_data = definition.func(**merge_data(path_data, request_data))
+            response_data = require_response_data(definition.func(**merge_data(path_data, request_data)))
```
adamhadani: @jessemyers  I noticed this decorator is now used for all the main CRUD operators, thought it should also be applied here for the 'For' versions?

jessemyers: That surprised me. I'd have only expected this to be present for DELETE where the return value of the func would be expected to indicate if the resource couldn't be found on delete.

microcosm_flask/conventions/relation.py
```diff
@@ -41,11 +42,56 @@ def configure_createfor(self, ns, definition):
...
+
+        """
+        @self.graph.route(ns.instance_path, Operation.DeleteFor, ns)
+        def delete(**path_data):
+            require_response_data(definition.func(**path_data))
+            return "", Operation.DeleteFor.value.default_code
+
+        delete.__doc__ = "Delete a {} relative to a {}".format(pluralize(ns.object_name), ns.subject_name)
+
+    def configure_replacefor(self, ns, definition):
```
jessemyers: I'd like to be convinced that this endpoint makes sense as a convention.

In my mind, we only use replace when we already have an object id (whether or not it exists in the database yet). I don't see why we'd operate on the relation in this case instead of the regular CRUD location.

jessemyers:
I'd like to discuss whether `ReplaceFor` is needed.

microcosm_flask/conventions/relation.py
adamhadani: @jessemyers you mean, once we have a concrete object id we can define relevant routes directly on it for things like Replace? 
That makes sense generally, however specifically for DynamoDB we ran into a situation where the DynamoDB API requires the composite key of (crawl_job_id, result_id) to access a 'result' (here crawl_job would be the subject, result would be the object of the relation). 
so the most straight-forward way of doing that seems to be to make it part of the URL scheme as in /api/crawl_job/:crawl_job_id/result/:id  - The alternative could be to require passing 'crawl_job_id' as a body field in request for /api/result/:result_id I suppose, but that seems much uglier. WDYT?

jessemyers: Is the crawl job its own resource?

If so, great, this makes sense, but please update the doc string to reflect when it makes sense to use this convention.

If not, I think the ideal answer here would actually be an extension of CRUD to have a composite id (`/crawl_job/:craw_job_id/:result_id`)

jessemyers:
LGTM if you resolve my last question

microcosm_flask/conventions/relation.py
adamhadani: @jessemyers yes crawl job is its own resource as well

microcosm_flask/conventions/relation.py
```diff
@@ -41,11 +42,62 @@ def configure_createfor(self, ns, definition):
...
+        """
+        Register a replace-for relation endpoint.
+
+        For typical usage, this relation is not strictly required; once an object exists and has its own ID,
+        it is better to operate on it directly via dedicated CRUD routes.
+        However, in some cases, the composite key of (subject_id, object_id) is required to look up the object.
+        This happens, for example, when using DynamoDB where an object which maintains both a hash key and a range key
+        requires specifying them both for access.
```
adamhadani: @jessemyers added pydoc

adamhadani:
@dumpstate @jessemyers  if you guys figured out we dont want this stuff LMK i'll close this PR as I no longer require it for the crawler purposes

jessemyers:
I think we want `DeleteFor`. I'm wary of `ReplaceFor`, but I don't think we reached a conclusion.

PR #39, opened 2016-06-09
Fixes for swagger with newer conventions:

jessemyers:
- Log common errors
 - Handle custom fields
 - Handle new relation operations

### DIFF
--- a/microcosm_flask/conventions/relation.py
+++ b/microcosm_flask/conventions/relation.py
@@ -64,7 +64,7 @@ class RelationConvention(Convention):
 
         @self.graph.route(ns.relation_path, Operation.RetrieveFor, ns)
         @qs(request_schema)
-        @response(definition.request_schema)
+        @response(definition.response_schema)
         def retrieve(**path_data):
             request_data = load_query_string_data(request_schema)
             response_data = definition.func(**merge_data(path_data, request_data))
--- a/microcosm_flask/conventions/swagger.py
+++ b/microcosm_flask/conventions/swagger.py
@@ -56,12 +56,15 @@ class SwaggerConvention(Convention):
     name="swagger",
     operations=[
         "create",
+        "create_for",
         "delete",
         "replace",
         "retrieve",
+        "retrieve_for",
         "search",
         "search_for",
         "update",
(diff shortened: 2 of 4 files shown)


microcosm_flask/conventions/relation.py
```diff
@@ -64,7 +64,7 @@ def configure_retrievefor(self, ns, definition):
 
         @self.graph.route(ns.relation_path, Operation.RetrieveFor, ns)
         @qs(request_schema)
-        @response(definition.request_schema)
+        @response(definition.response_schema)
```
jessemyers: This was just a typo/bug.

microcosm_flask/conventions/swagger.py
```diff
@@ -56,12 +56,15 @@ def discover():
     name="swagger",
```
jessemyers: We need these changes so that the new operations are mapped into the swagger doc.

microcosm_flask/swagger/schema.py
```diff
@@ -40,7 +45,18 @@ def build_parameter(field):
     See: https://github.com/marshmallow-code/apispec/blob/dev/apispec/ext/marshmallow/swagger.py#L81
 
     """
-    field_type, field_format = FIELD_MAPPINGS[type(field)]
+    try:
+        field_type, field_format = FIELD_MAPPINGS[type(field)]
+    except KeyError:
+        if hasattr(field, "__swagger_type__"):
+            field_type = getattr(field, "__swagger_type__")
```
jessemyers: I know you preferred the string fallback; I prefer exposing more control to the user. There will be custom non-string types at some point.

amitbn: do you really wanna leave the dependency on `__swagger_type__`?
Why not fallbacking the fallback with `str(field.__class__)`

amitbn:
+1

microcosm_flask/swagger/schema.py
jessemyers: I'd rather have this fail obviously (see: new exception logging) than assume the wrong type and have it fail cryptically somewhere downstream.

PR #41, opened 2016-06-17
Support null valued fields in swagger responses.

jessemyers:
Swagger apparently does not support null values; JSON Schema does, but nullability
needs to be expressed as a `oneOf` clause over both the `null` type and the actual
type of the object; Swagger appears to have eliminated this complexity with the
expectation that null values are just removed from the data type.

This change does a few things:

 - It supports pruning null values from microcosm JSON response; this behavior
   is conditional on the `X-Response-Skip-Null` header being supplied.

 - It injects this header as an optional parameter of every microcosm request.

 - It marks `allow_none` (marshmallow) fields as non-required Swagger fields.

From a Swagger client (e.g. Yelp's bravado), we can now handle null fields via:

    response = client.foo.bar(
        args,
        _request_options=dict(
            headers={
               "X-Response-Skip-Null": True,
            },
        ),
    )

### DIFF
--- a/microcosm_flask/swagger/definitions.py
+++ b/microcosm_flask/swagger/definitions.py
@@ -135,6 +135,19 @@ def body_param(schema):
     })
 
 
+def header_param(name, required=False, param_type="string"):
+    """
+    Build a header parameter definition.
+
+    """
+    return swagger.HeaderParameterSubSchema(**{
+        "name": name,
+        "in": "header",
+        "required": required,
+        "type": param_type,
+    })
+
+
 def query_param(name, field, required=False):
     """
     Build a query parameter definition.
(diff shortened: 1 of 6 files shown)


microcosm_flask/swagger/definitions.py
```diff
@@ -135,6 +135,19 @@ def body_param(schema):
     })
 
 
+def header_param(name, required=False, param_type="string"):
```
jessemyers: I tested out setting `required` to True for this header (because almost every API will return some null data so it's almost always correct to set this with a Swagger request). Unfortunately, I couldn't get bravado to validate when this happened; it's happy to send the header, but it doesn't connect the header it sends to the Swagger schema.

srt32:
LGTm

srt32:
LGTM

PR #42, opened 2016-06-28
Enable batch updates in sync-resources.

jessemyers:
It's useful to sync resources in batches.

### DIFF
--- a/microcosm_flask/conventions/crud_adapter.py
+++ b/microcosm_flask/conventions/crud_adapter.py
@@ -59,9 +59,19 @@ class CRUDStoreAdapter(object):
 
         """
         items = kwargs.pop("items")
+
+        def transform(item):
+            """
+            Transform the dictionary expected for replace (which uses the URI path's id)
+            into the resource expected from individual resources (which uses plain id).
+
+            """
+            item[self.identifier_key] = item.pop("id")
+            return item
+
         return dict(
             items=[
-                self.replace(**item)
+                self.replace(**transform(item))
                 for item in items
             ],
         )
--- a/microcosm_flask/sync/main.py
+++ b/microcosm_flask/sync/main.py
@@ -55,7 +55,7 @@ def parse_args():
         "-b",
         type=int,
         default=1,
-        help="Exclude the first resource from pull output",
+        help="Batch size (requires backend PATCH support if >1)",
     )
     parser.add_argument(
         "input",
--- a/microcosm_flask/sync/main.py
+++ b/microcosm_flask/sync/main.py
@@ -51,6 +51,13 @@ def parse_args():
         help="Exclude the first resource from pull output",
     )
     parser.add_argument(
+        "--batch-size",
+        "-b",
+        type=int,
+        default=1,
+        help="Exclude the first resource from pull output",
+    )
+    parser.add_argument(
         "input",
         help="Input location for resources",
     )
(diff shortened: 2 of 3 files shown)


microcosm_flask/conventions/crud_adapter.py
```diff
@@ -59,9 +59,19 @@ def update_batch(self, **kwargs):
 
         """
         items = kwargs.pop("items")
+
+        def transform(item):
```
jessemyers: After testing in an actual service, I discovered that we need this transform for a default/vanilla batch update.

microcosm_flask/sync/main.py
```diff
@@ -51,6 +51,13 @@ def parse_args():
         help="Exclude the first resource from pull output",
     )
     parser.add_argument(
+        "--batch-size",
+        "-b",
+        type=int,
+        default=1,
```
jessemyers: Default behavior is no batching.

microcosm_flask/sync/push.py
```diff
@@ -47,7 +50,25 @@ def push_json(inputs, base_url):
             scheme=parsed_base_url.scheme,
             netloc=parsed_base_url.netloc,
         ))
-        push_resource_json(uri, resource)
+
+        if batch_size == 1:
+            push_resource_json(uri, resource)
```
jessemyers: This is the current behavior.

microcosm_flask/conventions/crud_adapter.py
```diff
@@ -59,9 +59,19 @@ def update_batch(self, **kwargs):
...
+            """
+            Transform the dictionary expected for replace (which uses the URI path's id)
+            into the resource expected from individual resources (which uses plain id).
+
+            """
+            item[self.identifier_key] = item.pop("id")
```
srt32: What's an example of the mapping here?

microcosm_flask/sync/main.py
```diff
@@ -51,6 +51,13 @@ def parse_args():
...
+        help="Exclude the first resource from pull output",
```
srt32: Is this help text correct?

microcosm_flask/sync/push.py
```diff
@@ -47,7 +50,25 @@ def push_json(inputs, base_url):
...
+            continue
+
+        # batch handling
+        collection_uri = uri.rsplit("/", 1)[0]
+
+        if any((
+            current_uri is not None and current_uri != collection_uri,
```
srt32: What does this line check for?

srt32:
LGTM with some questions

microcosm_flask/conventions/crud_adapter.py
jessemyers: The individual resource functions all get called with an argument of the form `foo_id` because this is how we build our URI patterns (e.g. `/api/v1/foo/:foo_id`).

The resources themselves have an `id`; this translates between the two since we're patching with a list of resources and don't have individual URIs to extract the `foo_id` from.

microcosm_flask/sync/main.py
jessemyers: Definitely not.

microcosm_flask/sync/push.py
jessemyers: We either have an existing batch that is too large or we have a different resource (as interpreted by the collection URI). Either way, we need to submit the current batch before we go onto the next one.

In other words: we're NOT assuming the backend supports polymorphic updates with different resources in the same batch.

PR #43, opened 2016-06-29
Control whether sessions are used in sync-resources and handle connection errors

jessemyers:
I've been trying out using request sessions for batch updates to cache TCP connections. Unfortunately, I started seeing large numbers of connection resets, probably because some component in our system is not properly handling the full content read. (I've seen this before with uWSGI, but forgot where.)

As a result, I made `sync-resources` more tolerant of connection errors and make session usage opt-in.

### DIFF
--- a/microcosm_flask/sync/push.py
+++ b/microcosm_flask/sync/push.py
@@ -6,7 +6,8 @@ from json import dumps
 from logging import getLogger
 from sys import stdout
 
-from requests import Session
+import requests
+from requests.exceptions import ConnectionError
 from six.moves.urllib.parse import urlparse, urlunparse
 from yaml import safe_dump_all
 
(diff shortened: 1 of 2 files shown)


microcosm_flask/sync/push.py
```diff
@@ -25,16 +26,39 @@ def push_yaml(inputs, destination):
...
 
     Future implementations could perform paginated PATCH requests to a collection URI
     if our conventions support it.
 
     """
+    # either use a session or use plain requests
+    session_factory = requests.Session if enable_sessions else lambda: requests
+
+    session = session_factory()
+    for uri, resources in iter_json_batches(inputs, base_url, batch_size):
```
jessemyers: The batching logic is now a generator so we can focus on HTTP here.

microcosm_flask/sync/push.py
```diff
@@ -25,16 +26,39 @@ def push_yaml(inputs, destination):
...
+                else:
+                    push_resource_json_batch(session, uri, resources)
+            except ConnectionError as error:
+                logger.info("Connection error for uri: {}: {}".format(uri, error))
+                # on connection failure, recreate the session
+                session = session_factory()
+                continue
+            else:
+                break
+        if attempt == max_attempts:
```
jessemyers: In my tests, the connection resets always recovered after one failure. A down server failed the second time.

srt32:
LGtm

srt32:
LGTM

PR #44, opened 2016-06-29
Ensure that swagger works for multiple versions concurrently.

jessemyers:
The swagger endpoint was fixed, independent of version. This meant we couldn't have multiple versions.

### DIFF
--- a/microcosm_flask/namespaces.py
+++ b/microcosm_flask/namespaces.py
@@ -5,8 +5,11 @@ In conjunction with the `Operation` enum, defines a naming convention for HTTP e
 which in turn provides a discovery mechanism API routes.
 
 """
+from re import match
+
 from flask import request, url_for
 from six.moves.urllib.parse import urlencode, urljoin
+from werkzeug.exceptions import InternalServerError
 
 from microcosm_flask.naming import (
     collection_path_for,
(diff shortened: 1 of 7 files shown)


microcosm_flask/namespaces.py
```diff
@@ -95,30 +98,30 @@ def endpoint_for(self, operation):
...
-                self.subject_name,
-                operation.value.name,
-                self.object_name,
-            )
-        else:
-            return operation.value.pattern.format(
-                self.subject_name,
-                operation.value.name,
-            )
+        return operation.value.pattern.format(
```
jessemyers: I switched to keyword-based format strings so I can substitute without having to know which pattern is being used.

microcosm_flask/operations.py
```diff
@@ -27,6 +29,7 @@ class Operation(Enum):
     """
     # discovery operation
     Discover = OperationInfo("discover", "GET", NODE_PATTERN, 200)
+    DiscoverVersion = OperationInfo("discover_version", "GET", VERSIONED_NODE_PATTERN, 200)
```
jessemyers: Swagger now embeds the version into its endpoint pattern so we can have multiple at once with different versions.

microcosm_flask/operations.py
```diff
@@ -55,3 +58,15 @@ def from_name(cls, name):
...
         else:
             raise ValueError(name)
+
+    @property
+    def endpoint_pattern(self):
+        """
+        Convert the operation's pattern into a regex matcher.
+
+        """
+        parts = self.value.pattern.split(".")
```
jessemyers: This builds a matcher so we can extract kwargs back from the endpoint.

microcosm_flask/swagger/schema.py
```diff
@@ -38,6 +38,16 @@
 }
 
 
+SWAGGER_TYPE = "__swagger_type__"
+SWAGGER_FORMAT = "__swagger_format__"
+
+
+def swagger_field(field, swagger_type="string", swagger_format=None):
```
jessemyers: I added a wrapper function for a marshmallow field so we can more easily attach the customization attributes.

adamhadani:
LGTM

PR #46, opened 2016-07-05
Support passing context back into CRUDConvention.search

adamhadani:
This PR adds support for adding context into the default CRUDConvention endpoint.
This is useful, for example, when searching over nested resources, in which case we will want to supply
the root namespace's object_id.

Example usage: https://github.com/globality-corp/ursula/pull/17/files#diff-d4fec9dcac83546e70683d648eae36ffR76


jessemyers:
LGTM

PR #47, opened 2016-07-21
Don't merge in default credentials.

jessemyers:
If the user configures other credentials, do not enable the default testing credentials.


cdm-ium:
+1

PR #48, opened 2016-07-26
Version all endpoints

jessemyers:
This PR changes the internal endpoint names to always contain versions; otherwise we cannot easily have two of the same operation in different versions within the same application.

In a few places, I've injected `v1` as a default to help with backwards compatibility; in some cases, we'll have to start adding explicit versions. The most common of these will be in HAL-style links for existing resources in a non-v1 version.

Example... Was:

```
Link.for_(Operation.Retrieve, Foo, foo_id=foo_id)
```

Now:

```
Link.for_(Operation.Retrieve, Namespace(subject=Foo, version="v2"), foo_id=foo_id)
```

### DIFF
--- a/microcosm_flask/conventions/discovery.py
+++ b/microcosm_flask/conventions/discovery.py
@@ -56,7 +56,7 @@ class DiscoveryConvention(Convention):
         """
         page_schema = PageSchema()
 
-        @self.graph.route(ns.singleton_path, Operation.Discover, ns)
+        @self.graph.route("/", Operation.Discover, ns)
         def discover():
             # accept pagination limit from request
             page = Page.from_query_string(load_query_string_data(page_schema))
@@ -73,7 +73,7 @@ class DiscoveryConvention(Convention):
 
 
 @defaults(
-    name="",
+    name="hal",
     operations=[
         "search",
     ],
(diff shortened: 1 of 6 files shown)


microcosm_flask/conventions/discovery.py
```diff
@@ -73,7 +73,7 @@ def discover():
 
 
 @defaults(
-    name="",
+    name="hal",
```
jessemyers: I had to use a non-empty name here because Flask was very unhappy resoling the endpoint `.discovery.v1`. I kept the root path for compatibility.

microcosm_flask/conventions/swagger.py
```diff
@@ -45,7 +45,7 @@ def configure_discover(self, ns, definition):
         Register a swagger endpoint for a set of operations.
 
         """
-        @self.graph.route(ns.singleton_path, Operation.DiscoverVersion, ns)
```
jessemyers: We no longer need the versioned discovery operation; it's always versioned now.

microcosm_flask/conventions/discovery.py
netama: does this ever appear in a path?

netama:
LGTM

microcosm_flask/conventions/discovery.py
jessemyers: Not today.

PR #50, opened 2016-07-28
Added as_retryable exception function decorator

adamhadani:
Can be used around an exception to facilitate making it a retryable error when serializing into a transport (HTTP) error

### DIFF
--- a/microcosm_flask/errors.py
+++ b/microcosm_flask/errors.py
@@ -31,7 +31,7 @@ class ErrorSchema(Schema):
 def as_retryable(error):
     """
     Given an exception, mark it as retryable when serializing
-    into transporty layer error response.
+    into HTTP error response.
 
     """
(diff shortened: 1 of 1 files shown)


microcosm_flask/errors.py
```diff
@@ -28,6 +28,17 @@ class ErrorSchema(Schema):
     context = fields.Nested(ErrorContextSchema, required=False)
 
 
+def as_retryable(error):
+    """
+    Given an exception, mark it as retryable when serializing
+    into transporty layer error response.
+
+    """
+    error.retryable = True
```
jessemyers: Do you have a test case? I suspect this will work, but I know sometimes you need to use `setattr` if the underlying type isn't friendly to extension. I don't really know when that happens however.

jessemyers:
LGTM

PR #51, opened 2016-08-15
make it possible to log context with route requests

cdm-ium:
See:

- https://github.com/globality-corp/microcosm-flask/pull/51
- https://github.com/globality-corp/ariza/pull/22
- https://github.com/globality-corp/microcosm-logging/pull/15

My expectation is that things that want to use the context logger will define a `context_function` and pass it to a `context_logger` decorator. We can also start using this context function in place of manually parsing headers all over the place in code where we want to log/pass them to pubsub.

### DIFF
--- a/microcosm_flask/audit.py
+++ b/microcosm_flask/audit.py
@@ -17,23 +17,10 @@ from microcosm_flask.errors import (
 )
 
 
-X_REQUEST = "X-Request"
-
 AuditOptions = namedtuple("AuditOptions", [
     "include_request_body",
     "include_response_body",
-    "include_header",
 ])
-
-
-def include_header_matching_prefix(prefix):
-    """
-    Match all headers starting with prefix.
-
-    """
-    def include_header(header):
-        return header.startswith(prefix)
-    return include_header
 
 
 def audit(func):
@@ -48,14 +35,13 @@ def audit(func):
         options = AuditOptions(
             include_request_body=True,
             include_response_body=True,
-            include_header=include_header_matching_prefix(X_REQUEST),
         )
-        return _audit_request(options, func, *args, **kwargs)
+        return _audit_request(options, func, None, *args, **kwargs)
 
     return wrapper
 
 
-def _audit_request(options, func, *args, **kwargs):
+def _audit_request(options, func, request_context, *args, **kwargs):
     """
     Run a request function under audit.
 
@@ -81,11 +67,8 @@ def _audit_request(options, func, *args, **kwargs):
         audit_dict["request_body"] = request.get_json(force=True)
 
     # include headers (conditionally)
-    audit_dict.update({
-        header: value
-        for header, value in request.headers.items()
-        if options.include_header(header)
-    })
+    if request_context is not None:
+        audit_dict.update(request_context())
 
     # process the request
     try:
@@ -149,7 +132,6 @@ def parse_response(response):
 @defaults(
     include_request_body=True,
     include_response_body=True,
-    include_header_prefix=X_REQUEST,
 )
 def configure_audit_decorator(graph):
     """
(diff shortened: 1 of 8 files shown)


setup.py
```diff
@@ -42,6 +42,7 @@
             "app = microcosm_flask.factories:configure_flask_app",
             "audit = microcosm_flask.audit:configure_audit_decorator",
             "basic_auth = microcosm_flask.basic_auth:configure_basic_auth_decorator",
+            "context = microcosm_flask.context:configure_context",
```
jessemyers: Can we use a less general name?

jessemyers:
Looks sane except for naming. I'd also like a usage example... are there tests you can update to pass the controller around?

microcosm_flask/audit.py
```diff
@@ -17,25 +17,12 @@
 )
 
 
-X_REQUEST = "X-Request"
```
cdm-ium: I factored this code out into the context module.

setup.py
cdm-ium: How about `request_context`?

jessemyers: I think Flask already has one.

microcosm_flask/tests/test_namespaces.py
```diff
@@ -100,3 +101,24 @@ def search_foo():
     with graph.app.test_request_context():
         url = ns.href_for(Operation.Search)
         assert_that(url, is_(equal_to("http://localhost/api/foo")))
+
+
+def test_namespace_accepts_controller():
```
cdm-ium: Right now namespaces don't really do anything with the controller. It is just that I think it isn't entirely unreasonable for a namespace to declare who owns the implementation of routes for a given namespace.

We aren't doing this now, but we could use this to simplify the crud conventions to try to infer definitions from the namespace rather than explicitly registering them when configuring a controller, e.g.:

```
 def add_crud(graph, version):
     controller = graph.user_event_controller
      ns = Namespace(
          subject=UserEvent,
          controller=controller,
          version=version,
      )
      mappings = {
          Operation.Create: EndpointDefinition(
              func=transactional(controller.create),
             request_schema=NewUserEventSchema(),
             response_schema=UserEventSchema(),
         ),
         Operation.Delete: EndpointDefinition(
             func=transactional(controller.delete),
         ),
         Operation.Replace: EndpointDefinition(
             func=transactional(controller.replace),
             request_schema=NewUserEventSchema(),
             response_schema=UserEventSchema(),
         ),
         Operation.Retrieve: EndpointDefinition(
             func=controller.retrieve,
             response_schema=UserEventSchema(),
         ),
         Operation.Search: EndpointDefinition(
             func=controller.search,
             request_schema=UserEventSearchSchema(),
             response_schema=UserEventSchema(),
         ),
```

Could be:

```
 def add_crud(graph, version):
      controller = graph.user_event_controller
      ns = Namespace(
          subject=UserEvent,
          controller=controller,
          version=version,
      )
      mappings = {
          Operation.Create: EndpointDefinition(
             request_schema=NewUserEventSchema(),
             response_schema=UserEventSchema(),
         ),
         Operation.Delete: EndpointDefinition(
         ),
         Operation.Replace: EndpointDefinition(
             request_schema=NewUserEventSchema(),
             response_schema=UserEventSchema(),
         ),
         Operation.Retrieve: EndpointDefinition(
             response_schema=UserEventSchema(),
         ),
         Operation.Search: EndpointDefinition(
             request_schema=UserEventSearchSchema(),
             response_schema=UserEventSchema(),
         ),
```

For the sunny day case. Not sure how much it is worth, but pushing an instance of the controller to every definition seems redundant for something that should be constant per namespace.

jessemyers:
LGTM

PR #52, opened 2016-08-15
support adding basic auth to a specific namespace

netama:


microcosm_flask/namespaces.py
```diff
@@ -32,17 +32,19 @@ class Namespace(object):
 
     """
 
-    def __init__(self, subject, object_=None, path=None, version=None):
+    def __init__(self, subject, object_=None, path=None, version=None, enable_basic_auth=None):
```
jessemyers: Default to `False`?

jessemyers:
LGTM

microcosm_flask/namespaces.py
netama: I actually changed that intentionally but I guess it wasn't clear enough. I'll change the comment instead. Thanks

PR #56, opened 2016-08-16
Support allow_none fields to be null

srt32:
Why:

* we have some fields (such as a company's thumbnailUrl) that are of
  type string but allow_none True. We want to support `null`'s in this
  case but without an explicit allowance for this check, swagger clients
  will break when they recieve a null in place of a string.

This change addresses the need by:

* per
  https://github.com/OAI/OpenAPI-Specification/issues/229#issuecomment-193537548
  support of `null` as a string works as an additional type

### DIFF
--- a/microcosm_flask/swagger/schema.py
+++ b/microcosm_flask/swagger/schema.py
@@ -69,7 +69,10 @@ def build_parameter(field):
 
     parameter = {}
     if field_type:
-        parameter["type"] = field_type
+        if field.allow_none is True:
+            parameter["type"] = [field_type, 'null']
+        else:
+            parameter["type"] = field_type
     if field.metadata.get("description"):
         parameter["description"] = field.metadata["description"]
     if field_format:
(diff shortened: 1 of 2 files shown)


jessemyers:
LGTM

jessemyers:
You should definitely test this with a swagger client; I have a hunch that swagger prefers absent fields versus null values. I'm not sure that swagger has a `null` type that is valid.

srt32:
How does this idea mesh with https://github.com/globality-corp/microcosm-flask/blob/a1af7d6a514d91939202f78bdfeb7bf8960372c3/microcosm_flask/conventions/encoding.py#L72?

jessemyers:
Right. I remember that. I put some code in place so the `bravado` swagger client would work as long as it sent a header telling us it didn't null values. And then I made the swagger definitions set that header.

srt32:
Got it. For the record, something like this will work:

```
>>> retrieve = client.company.retrieve(
... company_id="1ff38a00-5f97-4fea-b175-fa799e99bac9",
... _request_options={"headers": {"X-Response-Skip-Null": "true"}}
... )
```

I will hold off on merging this PR until I can test further.

srt32:
Closing as the workaround in https://github.com/globality-corp/microcosm-flask/pull/56#issuecomment-240226718 is sufficient for the time being.

PR #57, opened 2016-08-26
adding opaque binding for flask routes

cdm-ium:
Flask application of: https://github.com/globality-corp/microcosm/pull/12

I can't merge this until I bump versions of microcosm.

### DIFF
--- a/microcosm_flask/routing.py
+++ b/microcosm_flask/routing.py
@@ -64,6 +64,9 @@ def configure_route_decorator(graph):
                     parent=ns.controller,
                 )
 
+            # set the opaque component data_func to look at the flask request context
+            func = graph.opaque.bind(graph.request_context)(func)
+
             # keep audit decoration last (before registering the route) so that
             # errors raised by other decorators are captured in the audit trail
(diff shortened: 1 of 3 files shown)


microcosm_flask/routing.py
```diff
@@ -40,6 +41,9 @@ def search_foo():
     """
     # routes depends on converters
     graph.use(*graph.config.route.converters)
+    # routes are annotated to make request context availale in
+    # graph.opaque
+    graph.use('opaque')
```
jessemyers: Consider moving this into `factories.py` instead; I only put the converters here because I was trying to support user-defined converters.

microcosm_flask/routing.py
```diff
@@ -8,6 +8,7 @@
 
 from microcosm.api import defaults
 from microcosm_logging.decorators import context_logger
+import microcosm.opaque  # noqa
```
jessemyers: This should be moved to `factories.py`

jessemyers:
LGTM with one file relocation

PR #58, opened 2016-08-26
add empty schema

netama:
Moved and renamed dummy schema from https://github.com/globality-corp/florentino/pull/264


jessemyers:
LGTM

PR #61, opened 2016-10-09
Correctly handle schema with no required fields in swagger definition.

jessemyers:
Previously, we were inserting a `required` attribute into every definition,
even if it had no required fields. This created problems with schema validation,
which expected the `required` value to be a non-empty list. Our workaround was
introducing the `EmptySchema` base with a placeholder required field.

It turns out that we can just omit the `required` field entirely if there are
no required fields. This makes swagger usage (and validation) much simpler.

### DIFF
--- a/microcosm_flask/generic_resources.py
+++ b/microcosm_flask/generic_resources.py
@@ -1,8 +1,0 @@
-from marshmallow import Schema, fields
-
-
-class EmptySchema(Schema):
-    """
-    Dummy schema for making empty requests without swagger validation errors.
-    """
-    _ = fields.Boolean(required=True, dump_only=True)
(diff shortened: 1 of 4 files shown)


microcosm_flask/tests/conventions/fixtures.py
```diff
@@ -38,6 +38,11 @@ class NewPersonBatchSchema(Schema):
     items = fields.List(fields.Nested(NewPersonSchema))
 
 
+class UpdatePersonSchema(Schema):
```
jessemyers: I added a test case that exercises this behavior.

microcosm_flask/generic_resources.py
```diff
@@ -1,8 +0,0 @@
-from marshmallow import Schema, fields
-
-
-class EmptySchema(Schema):
```
jessemyers: I'm removing this base; I will take care of removing its usage once this releases.

PR #62, opened 2016-10-11
Forward ports

jessemyers:
When our software is deployed behind an ALB, the client's port is not the port that Flask gets requested on. This PR adds an adapter that uses the `X-Forwarded-Port` to ensure that generated URIs (e.g. from `url_for`) reflect the client's view.

At the moment, we have no need for other forwarded headers; the scheme and host are appropriate.

### DIFF
--- a/microcosm_flask/forwarding.py
+++ b/microcosm_flask/forwarding.py
@@ -1,0 +1,46 @@
+"""
+Support for request forwarding.
+
+HATEOAS depends heavily on services being able to generate URIs back to their own resources.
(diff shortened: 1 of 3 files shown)


adamhadani:
LGTM

afallou:
Does the forwarding port change dynamically - meaning is it possible that it could change after the `url_for` has been generated but before the other service has fetched the resource through that url?

jessemyers:
No, this is the port that the client is currently using to make a connection to this service. The port that the service experience might be different, depending on the load-balancer, but the port can't change without starting a new request. (And in practice, we serve our APIs from well-known, fixed ports).

PR #63, opened 2016-10-12
Document url_for kwargs

takumig:


jessemyers:
LGTM

Note that `_external=True` is now the default behavior (changed since yesterday).

PR #64, opened 2016-10-12
clean null values recursively

netama:
`null` values in schemas are not supported. Extending the current `null` cleanup to remove `null` values recursively, and also adding it to the `search` responses.
Will try to improve the implementation at some point.


jessemyers:
LGTM

You might want to change the cleanup function name to `remove_null_values`.

PR #70, opened 2016-10-31
Support customized query string arguments during pagination

jessemyers:
If we have a filtered search operation, our paging doesn't automatically include the filter parameters we started with; this results in useless next/prev links.

This generalizes the paging to accept further arguments. It's a first pass with some obvious improvements, but I wanted feedback before finishing.

### DIFF
--- a/microcosm_flask/conventions/crud.py
+++ b/microcosm_flask/conventions/crud.py
@@ -21,6 +21,10 @@ from microcosm_flask.paging import Page, PaginatedList, make_paginated_list_sche
 
 class CRUDConvention(Convention):
 
+    @property
+    def page_cls(self):
+        return Page
+
     def configure_search(self, ns, definition):
         """
         Register a search endpoint.
@@ -43,7 +47,7 @@ class CRUDConvention(Convention):
         @response(paginated_list_schema)
         def search(**path_data):
             request_data = load_query_string_data(definition.request_schema)
-            page = Page.from_query_string(request_data)
+            page = self.page_cls.from_query_string(request_data)
             return_value = definition.func(**merge_data(path_data, request_data))
 
             if len(return_value) == 3:
--- a/microcosm_flask/conventions/crud.py
+++ b/microcosm_flask/conventions/crud.py
@@ -9,7 +9,6 @@ from microcosm_flask.conventions.encoding import (
     dump_response_data,
     load_query_string_data,
     load_request_data,
-    make_response,
     merge_data,
     require_response_data,
 )
@@ -56,9 +55,16 @@ class CRUDConvention(Convention):
                 context = {}
                 items, count = return_value
 
-            # TODO: use the schema for encoding
-            response_data = PaginatedList(ns, page, items, count, definition.response_schema, **context).to_dict()
-            return make_response(response_data)
+            response_data = PaginatedList(
+                ns=ns,
+                page=page,
+                items=items,
+                count=count,
+                schema=definition.response_schema,
+                operation=Operation.Search,
+                **context
+            )
+            return dump_response_data(paginated_list_schema, response_data)
 
         search.__doc__ = "Search the collection of all {}".format(pluralize(ns.subject_name))
 
--- a/microcosm_flask/conventions/relation.py
+++ b/microcosm_flask/conventions/relation.py
@@ -13,7 +13,6 @@ from microcosm_flask.conventions.encoding import (
     dump_response_data,
     load_query_string_data,
     load_request_data,
-    make_response,
     merge_data,
 )
 from microcosm_flask.conventions.registry import qs, request, response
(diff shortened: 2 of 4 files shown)


microcosm_flask/conventions/crud.py
```diff
@@ -21,6 +21,10 @@
 
 class CRUDConvention(Convention):
 
+    @property
```
jessemyers: I'm adding a hook here to choose a different pagination class. I'm hoping we don't actually need to use this hook.

microcosm_flask/paging.py
```diff
@@ -35,9 +35,11 @@ class PaginatedListSchema(Schema):
 
 
 class Page(object):
-    def __init__(self, offset, limit):
+
+    def __init__(self, offset, limit, **rest):
```
jessemyers: I've generalized the `Page` object to retain "other" arguments from its pagination. I have not yet generalized it's `from_query_string` function but I think I should. Feedback here please.

microcosm_flask/tests/test_paging.py
```diff
@@ -116,3 +116,56 @@ def search_foo():
                 },
             }
         })))
+
+
+def test_custom_paginated_list():
+    graph = create_object_graph(name="example", testing=True)
+    ns = Namespace(subject="foo", object_="bar")
+
+    class CustomPage(Page):
```
jessemyers: Here's an example of customized paging done generically. I'm contemplating moving this `@classmethod` as the default. This would mean that any arguments passed in the query string (as defined by the filter schema) would automatically end up in the pagination expansion. This is usually what we want when filtering down a search, but I was reluctant to make it the default behavior without some discussion.

microcosm_flask/tests/test_paging.py
```diff
@@ -116,3 +116,56 @@ def search_foo():
...
+                    "href": "http://localhost/api/foo/FOO_ID/bar?offset=2&limit=2&baz=baz",
+                },
+                "next": {
+                    "href": "http://localhost/api/foo/FOO_ID/bar?offset=4&limit=2&baz=baz",
+                },
+                "prev": {
+                    "href": "http://localhost/api/foo/FOO_ID/bar?offset=0&limit=2&baz=baz",
+                },
+            },
+            "baz": "baz",
```
jessemyers: The pagination code automatically injects all page arguments into the returned dictionary. This is ugly and doesn't validate against the paginated list schema that we [generate](https://github.com/globality-corp/microcosm-flask/blob/develop/microcosm_flask/paging.py#L25).

I can see a few ways to go here:
 - Hide these fields (easiest/most likely)
 - Actually use the paginated list schema to encode the data (better, more risky)
 - Generalize paging data into a sub dictionary (even better, not backwards compatible)
```
"paging": {
    "offset": 0,
    "limit": 20,
    "baz": "baz
}
``` 

I'm leaning towards the first...

jessemyers:
I pushed a second revision that addresses many of the concerns in the first revision:
 - `Page` itself is now more general
 - The pagination (search results) schema controls what gets encoded in the result
 - The paginated list resource still knows how to map itself to a dict, but this isn't really used anymore

microcosm_flask/conventions/crud.py
```diff
@@ -52,9 +55,16 @@ def search(**path_data):
...
+            response_data = PaginatedList(
+                ns=ns,
+                page=page,
+                items=items,
+                count=count,
+                schema=definition.response_schema,
+                operation=Operation.Search,
+                **context
+            )
+            return dump_response_data(paginated_list_schema, response_data)
```
jessemyers: No longer calling `to_dict()`; we end up building the response data the same as in other endpoints

microcosm_flask/conventions/relation.py
```diff
@@ -129,17 +128,16 @@ def search(**path_data):
...
+                page=page,
+                items=items,
+                count=count,
+                schema=definition.response_schema,
+                operation=Operation.SearchFor,
                 **context
-            ).to_dict()
-            return make_response(response_data)
+            )
+            return dump_response_data(paginated_list_schema, response_data)
```
jessemyers: Same logic for `SearchFor` as in `Search`

microcosm_flask/paging.py
```diff
@@ -47,21 +49,27 @@ def from_query_string(cls, qs):
...
         """
+        dct = qs.copy()
+        offset = dct.pop("offset")
+        limit = dct.pop("limit")
         return cls(
-            offset=qs["offset"],
-            limit=qs["limit"],
+            offset=offset,
+            limit=limit,
+            **dct
```
jessemyers: Every `Page` now has access to the `rest` of the query string arguments. By default, these will be shown in the paging URIs. (To omit, use a narrower paging query or sub-class `Page`)

microcosm_flask/paging.py
```diff
@@ -103,11 +114,23 @@ def to_dict(self):
                 self.schema.dump(item).data if self.schema else item
                 for item in self.items
             ],
-            _links=self.links.to_dict(),
+            _links=self._links,
```
jessemyers: The `to_dict()` method is basically deprecated. I haven't checked to see whether we use it anywhere else.

Instead, we're just passing this `PaginatedList` object to our schema. The main limitation here is that the values within `self.page.to_dict()` has to be accessed via `self.page.foo`; if we want these returned in the paging result, we need to subclass the paging schema and give it extra fields with these mappings.

netama:
LGTM

PR #72, opened 2016-11-04
Add QueryList field type

srt32:
Why:

* complement to
  https://github.com/globality-corp/oscar/tree/st-extract-query-list-params-field.
  This changeset introduces support for more flexible handling of query
  strings including lists of values.

This change addresses the need by:

* adding a new QueryStringList field type that handles parsing query
  strings of the following types:
** foo?foo_ids=a,b
** foo?foo_ids[]=a&foo_ids[]=b

Instead of using
`foo_ids = fields.List(fields.UUID, required=False)`

we can now user:

`foo_ids = QueryStringList(fields.UUID, required=False)`

### DIFF
--- a/microcosm_flask/fields.py
+++ b/microcosm_flask/fields.py
@@ -5,7 +5,7 @@ Custom fields.
 from datetime import datetime
 
 from dateutil import parser
-from marshmallow.fields import Field, ValidationError
+from marshmallow.fields import Field, List, ValidationError
 
 
 class EnumField(Field):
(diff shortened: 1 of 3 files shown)


microcosm_flask/fields.py
```diff
@@ -102,3 +102,27 @@ def _deserialize(self, value, attr, obj):
...
+        /foo?bars[]=1&bars[]2
+
+        and returns a list of values
+
+        """
+        if value is None:
+            return None
+
+        try:
+            attribute_elements = [attr_element.split(",") for attr_element in obj.getlist(attr)]
```
jessemyers: OMG, this looks so much better!

jessemyers:
LGTM

srt32:
Thanks!

PR #74, opened 2016-11-07
fix re-raising of last error on sync failure after all attempts to work in Python 3

adamhadani:
In Python 3, except error objects are cleared at end of their scope meaning the code as is raises a UnboundLocalError:
```UnboundLocalError: local variable 'error' referenced before assignment```

see for example: http://stackoverflow.com/questions/29268892/python-3-exception-deletes-variable-in-enclosing-scope-for-unknown-reason


jessemyers:
LGTM

Hooray for scoping

PR #75, opened 2016-11-07
Handle by_value and integer-valued enums in swagger conversion

jessemyers:
Many of our project mappings still need int enums.


netama:
LGTM

PR #76, opened 2016-11-08
Supporting keeping individual instance path URIs for bulk import

adamhadani:
Currently, the batch import mode for sync-resources (when using --batch-size) will default to truncate the individual instances URIs, assuming a top-level 'collection' resource to exist one level above individual instances URIs.

For cases where the instances themselves are however exported in batch and do not have a standard per-instance URI, e.g when exporting from a single /api/v1/data URI, this breaks.

This PR adds a simple flag to support keeping the individiual instance URIs (up to query parameter) when batch importing back into the service.

### DIFF
--- a/microcosm_flask/sync/push.py
+++ b/microcosm_flask/sync/push.py
@@ -26,7 +26,7 @@ def push_yaml(inputs, destination):
     safe_dump_all(({href: resource} for href, resource in inputs), destination)
 
 
-def push_json(inputs, base_url, batch_size, enable_sessions=False,  max_attempts=2):
+def push_json(inputs, base_url, batch_size, enable_sessions=False,  keep_instance_path=False, max_attempts=2):
     """
     Write inputs to remote URL as JSON.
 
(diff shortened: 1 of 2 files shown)


adamhadani:
this let’s us support all cases where there is a single ‘batch export’ URI (e.g all instances exported share a single URI / do not have meaningful self links). this works for the escolastica use case in this [PR](https://github.com/globality-corp/escolastica/pull/16) nicely.. open to suggestions on other ways to implement though

microcosm_flask/sync/push.py
```diff
@@ -91,7 +91,10 @@ def iter_json_batches(inputs, base_url, batch_size):
             yield (uri, [resource])
         else:
             # batch handling
-            collection_uri = uri.rsplit("/", 1)[0]
+            if keep_instance_path:
```
jessemyers: I wonder if we can find a way to express the batch spilt function on the CLI instead of having a conditional.

jessemyers:
LGTM

PR #77, opened 2016-11-08
Turn off logging of health endpoints

jessemyers:
Small exercise that will help reduce log spew.

### DIFF
--- a/microcosm_flask/audit.py
+++ b/microcosm_flask/audit.py
@@ -24,6 +24,26 @@ AuditOptions = namedtuple("AuditOptions", [
 ])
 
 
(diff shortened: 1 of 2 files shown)


microcosm_flask/audit.py
```diff
@@ -24,6 +24,26 @@
 ])
 
 
+SKIP_LOGGING = "_microcosm_flask_skip_audit_logging"
+
+
+def skip_logging(func):
```
jessemyers: This decorator can be used anywhere... but probably shouldn't be.

netama:
LGTM

PR #78, opened 2016-11-09
force strict validation for list resources

cdm-ium:
I wasted a lot of time debugging a validation error on a child resource of a list resource which failed to serialize properly and caused marshmallow to do weird things. We aren't using strict validation on our list resources which means when validation fails for an `item` on a paginated list, rather than raise an exception, marshmallow will just proceed like nothing happened and return an invalid resource.

### DIFF
--- a/microcosm_flask/conventions/crud.py
+++ b/microcosm_flask/conventions/crud.py
@@ -39,7 +39,7 @@ class CRUDConvention(Convention):
         :param definition: the endpoint definition
 
         """
-        paginated_list_schema = make_paginated_list_schema(ns, definition.response_schema)()
+        paginated_list_schema = make_paginated_list_schema(ns, definition.response_schema)(strict=True)
 
         @self.graph.route(ns.collection_path, Operation.Search, ns)
         @qs(definition.request_schema)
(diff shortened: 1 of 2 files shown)


microcosm_flask/conventions/crud.py
```diff
@@ -39,7 +39,7 @@ def configure_search(self, ns, definition):
         :param definition: the endpoint definition
 
         """
-        paginated_list_schema = make_paginated_list_schema(ns, definition.response_schema)()
+        paginated_list_schema = make_paginated_list_schema(ns, definition.response_schema)(strict=True)
```
cdm-ium: @jessemyers  is the best way to do it? It seems like adding a class properly on the class returned by `make_paginated_list_schema` won't do what I want.

jessemyers:
LGTM

Do you want to add a test that `strict` fixes your bug? Can we repro?

cdm-ium:
@jessemyers strict doesn't fix the bug, just gives a better stack trace, I can add a test.

jessemyers:
Unmerged since last November. Closing.

PR #79, opened 2016-11-10
default Content-Type header in response as application/json

adamhadani:
This adds setting default 'Content-Type' header for all responses as "application/json".
This is helpful especially when using things like Postman for manual API calls so that it uses the richer JSON view display for response


jessemyers:
LGTM

PR #80, opened 2016-11-11
Stringify enum members with .name

afallou:
Why:
Currently `search` endpoints don't treat `Enum` parameters well in the link they return. For example the `self` link to `https://service/api/v1/resource?something_type=BLAH` would be `https://service/api/v1/resource?something_type=SomethingType.BLAH`, which doesn't work

### DIFF
--- a/microcosm_flask/paging.py
+++ b/microcosm_flask/paging.py
@@ -2,6 +2,8 @@
 Pagination support.
 
 """
+from enum import Enum
+
 from marshmallow import fields, Schema
 
 from microcosm_flask.linking import Link, Links
@@ -84,9 +86,14 @@ class Page(object):
             ("offset", self.offset),
             ("limit", self.limit),
         ] + [
-            (key, self.rest[key])
+            (key, self._qs_value_for_tuple(self.rest[key]))
             for key in sorted(self.rest.keys())
         ]
+
+    def _qs_value_for_tuple(self, value):
+        if isinstance(value, Enum):
+            return value.name
+        return value
 
 
 class PaginatedList(object):


microcosm_flask/paging.py
```diff
@@ -84,10 +85,15 @@ def to_tuples(self):
             ("offset", self.offset),
             ("limit", self.limit),
         ] + [
-            (key, self.rest[key])
+            (key, self._qs_value_for_tuple(self.rest[key]))
             for key in sorted(self.rest.keys())
         ]
 
+    def _qs_value_for_tuple(self, value):
```
srt32: Is there a reasonable spec we could add? Perhaps we could update https://github.com/globality-corp/microcosm-flask/blob/d18311d4e88bdb1f0483d5bbc44c49c805af236d/microcosm_flask/tests/test_paging.py#L23?

srt32:
LGTM less a couple questions

microcosm_flask/paging.py
jessemyers: And I agree with @srt32 that we ought to have a test case that hits this.

netama: There are also `by_value` enums that should have `value.value` (such as https://github.com/globality-corp/florentino/blob/9c1a4b8a36800219e21e9a43855a2305e22145e9/florentino/resources/v1/qna_strategy_resources.py#L16)

afallou: Ok - will use `str()`, and the enums will have 
```
def __str__(self):
  return self.name
  # or, for by_value enums 
  return self.value
```

PR #81, opened 2016-12-01
Query String usage in Namespace.href_for

adamhadani:
This PR modifies the Namespace.href_for code to choose query string connector character when constructing namespace href by looking at whether other qs arguments are present.

### DIFF
--- a/microcosm_flask/tests/test_namespaces.py
+++ b/microcosm_flask/tests/test_namespaces.py
@@ -13,6 +13,7 @@ from mock import Mock
 from microcosm.api import create_object_graph
 from microcosm_flask.namespaces import Namespace
 from microcosm_flask.operations import Operation
+from microcosm_flask.tests.matchers import matches_url
 
 
 def test_endpoint_for():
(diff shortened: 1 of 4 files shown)


adamhadani:
I came across this issue when trying to build a Link like this:

```
       links["self"] = Link.for_(
            Operation.Retrieve,
            Namespace(
                subject=RDF_API_NAMESPACE,
                version="v1",
            ),
            offset=offset,
            limit=limit,
            qs=link_qs
        )
```

In this case, without this PR, the resulting URL looks like
http://localhost:5413/api/v1/rdf**?**limit=100&offset=100**?**type=http%3A%2F%2Fgraph.globality.io%2Fontology%2FGlobalityIndustry"

jessemyers:
LGTM; maybe add a unit test

microcosm_flask/tests/test_namespaces.py
```diff
@@ -120,6 +121,24 @@ def search_foo():
...
+    graph = create_object_graph(name="example", testing=True)
+    ns = Namespace(subject="foo")
+
+    @graph.route(ns.collection_path, Operation.Search, ns)
+    def search_foo():
+        pass
+
+    with graph.app.test_request_context():
+        url = ns.href_for(Operation.Search, offset=0, limit=10, qs=dict(foo="bar"))
+        assert_that(url, matches_url("http://localhost/api/foo?offset=0&limit=10&foo=bar"))
```
adamhadani: @jessemyers I realized when adding this unit-test that comparing URLs requires some canonicalization as otherwise the order of query strings is not guaranteed - I implemented as custom matcher using a third-party method I found. lmk if looks kosher to you, I imagine this matcher would be useful elsewhere also potentially

jessemyers: cc: @netama

PR #82, opened 2016-12-06
GLOB-5138: Add a URIField

jessemyers:
Includes:
 -  Reorganization of custom fields
 -  Addition of the `rfc3986` library and a URIField
 -  Elimination of the `urltools` library that makes an HTTP request to the Internet on load
 -  Movement of URIMatcher to an importable location


microcosm_flask/fields.py
```diff
@@ -1,128 +0,0 @@
-"""
-Custom fields.
```
jessemyers: Moved to a sub-package.

microcosm_flask/fields/uri_field.py
```diff
@@ -0,0 +1,93 @@
+"""
+A URI valued (and normalized) field.
```
jessemyers: This is new.

microcosm_flask/fields/uri_field.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+def normalize_uri_result(uri):
+    """
+    Normalize a URI (And return a URIResult)
+
+    """
+    ref = uri_reference(uri).normalize()
+
+    return ref._replace(
```
jessemyers: I'm still not super happy with the `rfc3986` library; it doesn't quite normalize as I'd hope so I wrote some extra processing below.

microcosm_flask/matchers.py
```diff
@@ -63,3 +67,30 @@ def schema_class(self):
 
     def describe_to(self, description):
         description.append_text("expected {}".format(prettify(self.expected)))
+
+
+class URIMatcher(BaseMatcher):
```
jessemyers: The previous `URLMatcher` (name changed  here) was in the `test` tree so it couldn't be imported in other projects. (We may not need to use it BTW if we normalize URIs on the input side via the new field)

microcosm_flask/tests/fields/test_uri_field.py
```diff
@@ -0,0 +1,79 @@
...
+from marshmallow import Schema, ValidationError
+from microcosm_flask.fields import URIField
+from microcosm_flask.fields.uri_field import normalize_uri
+
+
+class URISchema(Schema):
+    foo = URIField(include_query=True)
+
+
+def test_normalize_uri():
```
jessemyers: Test cases...

netama:
LGTM. Would adding another field type solve the schema validation error in https://dodona.dev.globality.io:5422/api/v1/swagger (for using a schema in the query)?

jessemyers:
@netama I'm not sure about that schema validation error.

PR #83, opened 2016-12-08
Log request resources when a batch import fails

adamhadani:
When using the batch import (PATCH) API, a failure will only log the URI, which is of little use.
This adds the current batch to the log - while it's pretty long, it's very useful for debugging.

Open to suggestions on better formatting etc.

### DIFF
--- a/microcosm_flask/sync/push.py
+++ b/microcosm_flask/sync/push.py
@@ -152,8 +152,9 @@ def push_resource_json_batch(session, uri, resources):
     try:
         response.raise_for_status()
     except:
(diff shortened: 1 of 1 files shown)


jessemyers:
My preference would be for you to extend the [request audit log](https://github.com/globality-corp/microcosm-flask/blob/develop/microcosm_flask/audit.py) so that your specific endpoint does this rather than making this the default behavior.

Otherwise, we risk leaking sensitive data by accident.

There's already some support in the audit log for switching on `Flask.g` per request.

PR #85, opened 2017-01-18
Improve support for allow_none

pokey:
Switch to using `'x-nullable': true` for marshmallow fields that have `allow_none=True`.  This field is not part of the official spec, but seems to be the currently accepted workaround, and it is supported by [bravado](https://github.com/Yelp/bravado-core/).

### DIFF
--- a/microcosm_flask/swagger/schema.py
+++ b/microcosm_flask/swagger/schema.py
@@ -122,6 +122,12 @@ def build_parameter(field):
     if isinstance(field, fields.List):
         parameter["items"] = build_parameter(field.container)
 
+    if field.allow_none:
+        # XXX Update to new name when we switch to next version of OpenAPI
+        # spec
+        # See https://github.com/OAI/OpenAPI-Specification/pull/741
+        parameter["x-nullable"] = True
+
     return parameter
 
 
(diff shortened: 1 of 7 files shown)


pokey:
Would be great to get feedback from @srt32, @jessemyers or @netama on this

microcosm_flask/swagger/schema.py
```diff
@@ -134,7 +140,7 @@ def build_schema(marshmallow_schema):
     required_fields = [
         field.dump_to or name
         for name, field in fields
-        if field.required and not field.allow_none
```
netama: I think this may break some existing calls in services (even though this looks like the right fix), so maybe notify people when publishing the new package?

netama:
LGTM if you tested this with some marquez calls

pokey:
@netama I've tested with babilonia and escolastica, but not any others.  I believe these are the only repos I have access to

microcosm_flask/swagger/schema.py
pokey: Ok.  How should I notify people?

netama: slack? (we can also do that)

jessemyers:
LGTM

adamhadani:
Reverted this change as encountered showstopper breakage due to https://github.com/Yelp/bravado-core/pull/143

PR #86, opened 2017-01-18
Hack to fix error reporting

pokey:
When an error is raised, microcosm flask tries to serialize the error context to json to send to the client.  However, the error context cannot be serialized to json.  This hack just converts the error context to a string.  This at least allows the server to send something coherent in the case of an error, but there are probably a lot better ways to handle this, potentially even by just getting rid of the context.  Another option would be to serialize its stack trace instead.


jessemyers:
Very old. Closing.

PR #96, opened 2017-02-01
Add Alias operation to redirect to a resource Retrieve endpoint

jessemyers:
This convention allows for "vanity" URLs that alias UUID-based resources using a human-
readable name (as long as the name is valid as part of the URL path).

A mapping for an Alias operation supports a `GET` based request with a path parameter
of the form `foo_name`. The mapping function is expected to return a valid resource
(or raise `NotFound`); the convention will issue a redirect to the target resource.

### DIFF
--- a/microcosm_flask/conventions/alias.py
+++ b/microcosm_flask/conventions/alias.py
@@ -1,0 +1,47 @@
+"""
+Conventions for aliasing endpoints.
+
+"""
+from flask import redirect
+from microcosm_flask.conventions.base import Convention
+from microcosm_flask.naming import name_for
+from microcosm_flask.operations import Operation
+
+
+class AliasConvention(Convention):
+
+    def configure_alias(self, ns, definition):
+        """
+        Register an alias endpoint which will redirect to a resource's retrieve endpoint.
+
+        Note that the retrieve endpoint MUST be registered prior to the alias endpoint.
+
+        The definition's func should be a retrieve function, which must:
+        - accept kwargs for path data
+        - return a resource
+
+        :param ns: the namespace
+        :param definition: the endpoint definition
+
+        """
+        @self.graph.route(ns.alias_path, Operation.Alias, ns)
+        def retrieve(**path_data):
+            resource = definition.func(**path_data)
+
+            kwargs = dict()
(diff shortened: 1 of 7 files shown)


microcosm_flask/conventions/alias.py
```diff
@@ -0,0 +1,47 @@
...
+from microcosm_flask.operations import Operation
+
+
+class AliasConvention(Convention):
+
+    def configure_alias(self, ns, definition):
+        """
+        Register an alias endpoint which will redirect to a resource's retrieve endpoint.
+
+        Note that the retrieve endpoint MUST be registered prior to the alias endpoint.
```
jessemyers: There are two reasons for this condition:
 1. The value of `ns.alias_path` (used by Alias) will shadow the value of `ns.instance_path` (used by Retrieve); if we get the order wrong, Flask will probably invoke the wrong function.
 2. The implementation on L34 looks up the Retrieve endpoint and resolves its URL. So it must exist.

microcosm_flask/swagger/definitions.py
```diff
@@ -121,11 +122,13 @@ def build_path(operation, ns):
     try:
         return ns.url_for(operation, _external=False)
     except BuildError as error:
+        # we are missing some URI path parameters
         uri_templates = {
             argument: "{{{}}}".format(argument)
             for argument in error.suggested.arguments
         }
-        return ns.url_for(operation, _external=False, **uri_templates)
+        # flask will sometimes try to quote '{' and '}' characters
```
jessemyers: I'm not enabling swagger definitions for Alias endpoints by default, but they will work to a first order as long as we use `unquote` here. Otherwise, the substitution that happens here will break for some obscure Flask reason.

NB: the path for an Alias to Foo is `/foo/{foo_name}`, but Flask's `url_for()` expects us to substitute a *real* value for `foo_name` into this path. So we substitute the string literal `"{foo_name}"`... this works except somehow if the literal doesn't contain the usual "uuid:" qualifier, Flask decides to escape the curly braces.

Ugh.

adamhadani:
LGTM

netama:
LGTM. I think for our use case we might also need to add version (not sure how to generalize that though)

PR #97, opened 2017-02-02
Allow the default pagination limit to be set by the requester.

jessemyers:
There doesn't seem to be any standard way to control the pagination limit
from the client side. The `Range` header is promising, but seems geared
toward bytes-oriented and token-based pagination. Invention a custom range
type doesn't seem better than using a custom header.

### DIFF
--- a/microcosm_flask/paging.py
+++ b/microcosm_flask/paging.py
@@ -48,7 +48,10 @@ class Page(object):
 
     @property
     def default_limit(self):
-        return request.headers.get("X-Request-Limit", 20)
+        try:
+            return int(request.headers["X-Request-Limit"])
+        except:
+            return 20
 
     @classmethod
     def from_query_string(cls, qs):
--- a/microcosm_flask/paging.py
+++ b/microcosm_flask/paging.py
@@ -2,6 +2,7 @@
 Pagination support.
 
 """
+from flask import request
 from marshmallow import fields, Schema
 
 from microcosm_flask.linking import Link, Links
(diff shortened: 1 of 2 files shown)


microcosm_flask/paging.py
```diff
@@ -2,15 +2,16 @@
...
 from marshmallow import fields, Schema
 
 from microcosm_flask.linking import Link, Links
 from microcosm_flask.operations import Operation
 
 
 class PageSchema(Schema):
-    offset = fields.Integer(missing=0, default=0)
-    limit = fields.Integer(missing=20, limit=20)
+    offset = fields.Integer(missing=None)
```
jessemyers: Stop setting defaults globally.

microcosm_flask/paging.py
```diff
@@ -36,11 +37,19 @@ class PaginatedListSchema(Schema):
 
 class Page(object):
 
-    def __init__(self, offset, limit, **rest):
-        self.offset = offset
-        self.limit = limit
+    def __init__(self, offset=None, limit=None, **rest):
+        self.offset = self.default_offset if offset is None else offset
```
jessemyers: Set the defaults as part of the `Page` class.

microcosm_flask/paging.py
```diff
@@ -36,11 +37,19 @@ class PaginatedListSchema(Schema):
...
+        self.limit = self.default_limit if limit is None else limit
         self.rest = rest
 
+    @property
+    def default_offset(self):
+        return 0
+
+    @property
+    def default_limit(self):
+        return request.headers.get("X-Request-Limit", 20)
```
jessemyers: Allow `X-Request-Limit` header to control the default limit

microcosm_flask/paging.py
```diff
@@ -36,11 +37,22 @@ class PaginatedListSchema(Schema):
...
         self.rest = rest
 
+    @property
+    def default_offset(self):
+        return 0
+
+    @property
+    def default_limit(self):
+        try:
+            return int(request.headers["X-Request-Limit"])
```
jessemyers: Allow X-Request-Limit header to control the default limit

PR #98, opened 2017-02-03
Add decorators to help manage what fields we log in our audit logs.

jessemyers:
Note that we currently suppress all request/response data in our logs
unless we're running in debug. The next step here would be to revisit
the audit log code to allow *some* request/response data based on this
or similar logic.


netama:
+1

PR #99, opened 2017-02-04
Fix offset and limit defaults for page resources.

cdm-ium:
Why? In: 6b31a5228339324c9d392de97f7ab8b5f1e60508 we moved offset and
limit defaulting from the PageSchema to the Page class, but we
don't call it properly when deserializing query results. This PR
leverages the Page resource to get query param defaulting.

### DIFF
--- a/microcosm_flask/conventions/crud.py
+++ b/microcosm_flask/conventions/crud.py
@@ -47,7 +47,7 @@ class CRUDConvention(Convention):
         def search(**path_data):
             request_data = load_query_string_data(definition.request_schema)
             page = self.page_cls.from_query_string(request_data)
-            return_value = definition.func(**merge_data(path_data, request_data))
+            return_value = definition.func(**merge_data(path_data, page.to_dict()))
 
             if len(return_value) == 3:
                 items, count, context = return_value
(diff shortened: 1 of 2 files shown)


microcosm_flask/conventions/crud.py
```diff
@@ -47,7 +47,7 @@ def configure_search(self, ns, definition):
         def search(**path_data):
             request_data = load_query_string_data(definition.request_schema)
             page = self.page_cls.from_query_string(request_data)
-            return_value = definition.func(**merge_data(path_data, request_data))
+            return_value = definition.func(**merge_data(path_data, page.to_dict()))
```
cdm-ium: Before my change, `crud` would just pass `load_query_string_data(definition.request_schema)` to the definition function by bypassing `Page` altogether. This is effectively equivalent to `Page.from_query_string(PageSchema()).load(request.args).data)`.

Previously we moved the offset and limit defaulting to `self.page_cls` (which is a `Page` instance) but never tied the two together so the defaulting of `offset` and `limit` went away.

jessemyers:
LGTM

Thanks for finding this.

PR #100, opened 2017-02-06
remove string cast from Page serialization

adamhadani:
Fixes [this issue](https://github.com/globality-corp/dodona/pull/22#issuecomment-277785997)

### DIFF
--- a/microcosm_flask/paging.py
+++ b/microcosm_flask/paging.py
@@ -2,6 +2,9 @@
 Pagination support.
 
 """
+from enum import Enum
+from uuid import UUID
+
 from flask import request
 from marshmallow import fields, Schema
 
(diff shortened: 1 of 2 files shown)


jessemyers:
LGTM

jessemyers:
LGTM even better

microcosm_flask/paging.py
```diff
@@ -96,10 +99,25 @@ def to_tuples(self):
...
+
+        :param value - a query parameter value
+        :returns value cast to string if value is one of pre-specified types (UUID, Enum)
+            or the value as-is otherwise.
+
+        """
+        if isinstance(value, (UUID, Enum)):
+            return str(value)
+
+        return value
```
adamhadani: ping @jessemyers I ended up adding this `to_value` method as a unit-test was failing here when trying to pass enum and UUID objects into the `from_query_string` factory and expect to see them serialized properly into a JSON response (as strings). 

1. I'm not sure this is a case we care about so perhaps unit-test is off?

2. In any case, I think Enum/UUID covers most mainstream use-cases for non offset/limit parameters so this would preserve backward-compatibility pretty well. removing this might break a bunch of services in hard to debug ways.. WDYT?

PR #102, opened 2017-02-08
Fix paging attempt #2

adamhadani:
This is an attempt to fix paging following recent regressions.

Roughly, my understanding of whats happening is something like:

* We moved bunch of code (specifically defaulting offset/limit values) from PageSchema to Paging
* This involved changing code to use paging.to_dict() instead of raw request_data in the search API convention
* This had undesired side-effect since Paging.to_dict casts everything to string - this cast is valid in context of generating JSON response, but not in context of passing the parsed request data to the Search method
* My previous PR #100 did not account for this dual usage

This PR attempts to solve the issue by explicitly forking the usage using a new `as_str` parameter that can be passed to Paging.to_dict() - it's set to True for when using as part of PagingList to serialize results, and set to False when using for parsing results and passing to `search` method.

See also: https://globalityinc.slack.com/archives/dev-backend/p1486572906001121

### DIFF
--- a/microcosm_flask/conventions/crud.py
+++ b/microcosm_flask/conventions/crud.py
@@ -47,7 +47,7 @@ class CRUDConvention(Convention):
         def search(**path_data):
             request_data = load_query_string_data(definition.request_schema)
             page = self.page_cls.from_query_string(request_data)
-            return_value = definition.func(**merge_data(path_data, page.to_dict()))
+            return_value = definition.func(**merge_data(path_data, page.to_dict(as_str=False)))
 
             if len(return_value) == 3:
                 items, count, context = return_value
--- a/microcosm_flask/paging.py
+++ b/microcosm_flask/paging.py
@@ -2,14 +2,19 @@
 Pagination support.
 
 """
-from enum import Enum
-from uuid import UUID
-
 from flask import request
 from marshmallow import fields, Schema
 
 from microcosm_flask.linking import Link, Links
 from microcosm_flask.operations import Operation
+
+
+def identity(x):
+    """
+    Identity function.
+
+    """
+    return x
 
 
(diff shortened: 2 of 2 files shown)


microcosm_flask/conventions/crud.py
```diff
@@ -47,7 +47,7 @@ def configure_search(self, ns, definition):
         def search(**path_data):
             request_data = load_query_string_data(definition.request_schema)
             page = self.page_cls.from_query_string(request_data)
-            return_value = definition.func(**merge_data(path_data, page.to_dict()))
+            return_value = definition.func(**merge_data(path_data, page.to_dict(as_str=False)))
```
adamhadani: When using this to pass to API method, do not do any casting

microcosm_flask/paging.py
```diff
@@ -145,7 +137,7 @@ def to_dict(self):
                 for item in self.items
             ],
             _links=self._links,
-            **self.page.to_dict()
+            **self.page.to_dict(as_str=True)
```
adamhadani: When using this as part of PaginatedList to serialize results to JSON repsonse,  cast to string

jessemyers:
I think we're getting further away from what we want, which is to use marshmallow for the conversions of the data we pass to our search function. I was hoping there'd be a way to encapsulate the page/limit defaulting logic so it can be used along with the regular marshmallow processing, though I'm not convinced that can be done cleanly.

adamhadani:
@jessemyers well I think it's getting closer because this will undo the cast everything to string part :) but maybe the correct solution involves `definition.request_schema` somehow?

netama:
LGTM, though I would hope to revert back to using `request_data` instead of `page.x`

PR #104, opened 2017-02-15
Add a count convention

jessemyers:
We often wish to query the number of things without returning results. While we can do
this easily with a `Search` and `limit=0`, it's also nice to use a dedicated endpoint.

Changes:
 -  Allow `HEAD /api/collection`
 -  Inject an `X-Total-Count` header for all search and count responses

To use, we just need to add something like:

```
        Operation.Count: EndpointDefinition(
            func=controller.count,
            request_schema=SearchFooSchema(),
        ),
```

### DIFF
--- a/microcosm_flask/conventions/crud_adapter.py
+++ b/microcosm_flask/conventions/crud_adapter.py
@@ -42,6 +42,10 @@ class CRUDStoreAdapter(object):
         count = self.store.count(**kwargs)
         return items, count
 
+    def count(self, offset=None, limit=None, **kwargs):
+        count = self.store.count(**kwargs)
+        return count
+
     def update(self, **kwargs):
         identifier = kwargs.pop(self.identifier_key)
         model = self.store.model_class(id=identifier, **kwargs)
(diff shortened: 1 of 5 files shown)


microcosm_flask/conventions/crud_adapter.py
```diff
@@ -42,6 +42,10 @@ def search(self, offset, limit, **kwargs):
         count = self.store.count(**kwargs)
         return items, count
 
+    def count(self, offset=None, limit=None, **kwargs):
```
jessemyers: I added `offset` and `limit` with default values here to ensure they are never passed to the count function. This allows us to reuse search schemas for count queries.

microcosm_flask/conventions/encoding.py
```diff
@@ -17,6 +17,12 @@ def with_context(error, errors):
     return error
 
 
+def encode_count_header(count):
+    return {
+        "X-Total-Count": count,
```
jessemyers: This header seemed semi-standard.

afallou:
LGTM - would it make sense to automatically make the `count` endpoint accessible when a search endpoint is defined?

jessemyers:
None of the other conventions are automatic, so I'd rather not start doing that here.

I could see a higher level abstraction for "standard crud" that takes a controller and 3 schemas and does the rest.

PR #105, opened 2017-02-24
Add upload conventions

jessemyers:
-  Supports an upload handler
 -  Sets consumes media type in swagger
 -  Assumes no response (204)

### DIFF
--- a/microcosm_flask/conventions/upload.py
+++ b/microcosm_flask/conventions/upload.py
@@ -1,0 +1,44 @@
+"""
+Conventions for file upload.
+
+"""
+from flask import request
+from microcosm_flask.conventions.base import Convention
+from microcosm_flask.operations import Operation
+from werkzeug.exceptions import BadRequest
+
+
+class UploadConvention(Convention):
+
+    def configure_upload(self, ns, definition):
+        """
+        Register an upload endpoint.
+
+        The definition's func should be an upload function, which must:
+        - accept kwargs for path data
+        - accept a dict of files
+
+        :param ns: the namespace
+        :param definition: the endpoint definition
+
+        """
+        @self.graph.route(ns.collection_path, Operation.Upload, ns)
+        def upload(**path_data):
+            files = request.files
+
+            if not files:
+                raise BadRequest("No files uploaded!")
+
+            definition.func(files, **path_data)
+            return "", Operation.Upload.value.default_code
+
+        upload.__doc__ = "Upload a file for: {}".format(ns.subject_name)
+
+
+def configure_upload(graph, ns, mappings):
+    """
+    Register Upload endpoints for a resource object.
+
+    """
+    convention = UploadConvention(graph)
+    convention.configure(ns, mappings)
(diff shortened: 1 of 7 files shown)


microcosm_flask/conventions/upload.py
```diff
@@ -0,0 +1,44 @@
...
+
+        The definition's func should be an upload function, which must:
+        - accept kwargs for path data
+        - accept a dict of files
+
+        :param ns: the namespace
+        :param definition: the endpoint definition
+
+        """
+        @self.graph.route(ns.collection_path, Operation.Upload, ns)
```
jessemyers: Limitation #1: This is using the "collection" path (although it could also be the "singleton" path). That means we're not uploading in the context of something else. For that, we'd need an `UploadFor` convention.

microcosm_flask/conventions/upload.py
```diff
@@ -0,0 +1,44 @@
...
+        def upload(**path_data):
+            files = request.files
+
+            if not files:
+                raise BadRequest("No files uploaded!")
+
+            definition.func(files, **path_data)
+            return "", Operation.Upload.value.default_code
```
jessemyers: Limitation 2: This is not returning anything. We could support a return value from `func` and an optional schema , but it comes at the cost of making the swagger definition smarter about multiple response types.

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py27, py35, lint
+envlist = py27, py36, lint
```
jessemyers: Homebrew is now shipping `3.6`, so I upgraded this too.

adamhadani:
LGTM

microcosm_flask/conventions/upload.py
```diff
@@ -0,0 +1,44 @@
...
+
+        """
+        @self.graph.route(ns.collection_path, Operation.Upload, ns)
+        def upload(**path_data):
+            files = request.files
+
+            if not files:
+                raise BadRequest("No files uploaded!")
+
+            definition.func(files, **path_data)
```
adamhadani: @jessemyers FYI there is a bit of nasty boilerplate further required to actually use an uploaded file - see [this](https://github.com/globality-corp/escolastica/blob/develop/escolastica/utils/file_utils.py#L11). Would it make sense to move this 'upload_tmp_file' helper to microcosm-flask somewhere?

jessemyers: Can you elaborate? Why is a temporary file needed vs just reading the input data?

adamhadani: It looks like above a certain fixed threshold (500Kb?) Flask will not make the file data available for streaming in memory from `file.read()` without saving it to temporary location first - http://stackoverflow.com/a/42405115
might be fine for most use cases to assume this size but in case does go over, the method linked to above uses flask's recommended secure_filename for handling saving it

microcosm_flask/conventions/upload.py
wheremiah: I'll file a ticket for that, I am guessing it will be useful for the crespi work, and I've been looking for an opportunity to add something to microcosm.

jessemyers: I might get there first...

jessemyers:
I'm going to close and reopen this with fixes.

PR #106, opened 2017-02-24
Add Upload and UploadFor operations and conventions

jessemyers:
-  Supports an upload handler
 -  Saves files safely to a temporary locaiton
 -  Sets consumes media type in swagger
 -  Supports query string and path arguments
 -  Supports no content or resource responses

### DIFF
--- a/microcosm_flask/conventions/upload.py
+++ b/microcosm_flask/conventions/upload.py
@@ -54,8 +54,10 @@ def temporary_upload(name, fileobj):
     filename = secure_filename(fileobj.filename)
     filepath = join(tempdir, filename)
     fileobj.save(filepath)
-    yield name, filepath, fileobj.filename
-    rmtree(tempdir)
+    try:
+        yield name, filepath, fileobj.filename
+    except:
+        rmtree(tempdir)
 
 
 class UploadConvention(Convention):
--- a/microcosm_flask/conventions/upload.py
+++ b/microcosm_flask/conventions/upload.py
@@ -2,7 +2,7 @@
 Conventions for file upload.
 
 """
-from contextlib import contextmanager, nested
+from contextlib import contextmanager
 from os.path import join
 from shutil import rmtree
 from tempfile import mkdtemp
@@ -19,6 +19,24 @@ from microcosm_flask.conventions.registry import qs, response
 from microcosm_flask.operations import Operation
 from werkzeug.exceptions import BadRequest
 from werkzeug.utils import secure_filename
+
+
+try:
+    # Python 2
+    from contextlib import nested
+except ImportError:
+    # Python3
+    from contextlib import ExitStack
+
+    @contextmanager
+    def nested(*contexts):
+        """
+        Reimplementation of nested in python 3.
+        """
+        with ExitStack() as stack:
+            for ctx in contexts:
+                stack.enter_context(ctx)
+            yield contexts
 
 
 @contextmanager
--- a/microcosm_flask/conventions/upload.py
+++ b/microcosm_flask/conventions/upload.py
@@ -1,0 +1,116 @@
+"""
+Conventions for file upload.
+
+"""
+from contextlib import contextmanager, nested
+from os.path import join
+from shutil import rmtree
+from tempfile import mkdtemp
+
+from flask import request
+from marshmallow import Schema
+from microcosm_flask.conventions.base import Convention
+from microcosm_flask.conventions.encoding import (
+    dump_response_data,
+    load_query_string_data,
+    merge_data,
+)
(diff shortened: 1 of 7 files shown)


microcosm_flask/conventions/upload.py
```diff
@@ -0,0 +1,134 @@
...
+    load_query_string_data,
+    merge_data,
+)
+from microcosm_flask.conventions.registry import qs, response
+from microcosm_flask.operations import Operation
+from werkzeug.exceptions import BadRequest
+from werkzeug.utils import secure_filename
+
+
+try:
```
jessemyers: Python3 went a different direction with nested context managers.

microcosm_flask/conventions/upload.py
```diff
@@ -0,0 +1,136 @@
...
+            request_data = load_query_string_data(request_schema)
+
+            if not request.files:
+                raise BadRequest("No files were uploaded")
+
+            uploads = [
+                temporary_upload(name, fileobj)
+                for name, fileobj
+                in request.files.items()
+                if not self.exclude_func(name, fileobj)
```
jessemyers: We can write custom filter to exclude specific files (e.g. based on extension)

microcosm_flask/conventions/upload.py
```diff
@@ -0,0 +1,136 @@
...
+            ]
+            with nested(*uploads) as files:
+                response_data = definition.func(files, **merge_data(path_data, request_data))
```
jessemyers: We pass a list of tuples containing the multipart form's field name, the uploaded file path, and the uploaded file's original name.

We also pass the path and query string data.

microcosm_flask/swagger/definitions.py
```diff
@@ -253,6 +259,10 @@ def add_responses(swagger_operation, operation, ns, func):
         if not hasattr(swagger_operation, "produces"):
             swagger_operation.produces = []
         swagger_operation.produces.append(response_resource)
+    elif not response_resource:
```
jessemyers: I had to add some customization for 200 vs 204 responses. In order to allow the success case of an operation to have a variable status code, I can't just use `operation.value.default_code`. Rather than generalize the operation to code mapping with some new function, I'm just handling the 204 case.

microcosm_flask/tests/conventions/test_upload.py
```diff
@@ -0,0 +1,169 @@
...
+
+class FileExtraSchema(Schema):
+    extra = fields.String(missing="something")
+
+
+class FileResponseSchema(Schema):
+    id = fields.UUID(required=True)
+
+
+UPLOAD_MAPPINGS = {
```
jessemyers: I have examples of:
 - `Update` and `UpdateFor`
 - 204 and 200 responses
 - query string parameters

microcosm_flask/tests/conventions/test_upload.py
```diff
@@ -0,0 +1,169 @@
...
+            ),
+        )
+        assert_that(
+            upload_for["parameters"],
+            has_item(
+                is_not(has_entry("name", "extra")),
+            ),
+        )
+
+        # one endpoint gets a custom response type (and the other doesn't)
```
jessemyers: TIL I learn about chaining matchers!

microcosm_flask/conventions/upload.py
```diff
@@ -0,0 +1,136 @@
...
+            ]
+            yield results
+
+
+@contextmanager
+def temporary_upload(name, fileobj):
+    """
+    Upload a file to a temporary location.
+
+    Flask will not load sufficiently large (>500M) files into memory, so it
```
adamhadani: only reference for limit I could find was [this](http://stackoverflow.com/a/15782516/1152258) which stated 500KB rather than 500M

jessemyers: got it

microcosm_flask/conventions/upload.py
```diff
@@ -0,0 +1,136 @@
...
+    makes sense to always load files into a temporary directory.
+
+    """
+    tempdir = mkdtemp()
+    filename = secure_filename(fileobj.filename)
+    filepath = join(tempdir, filename)
+    fileobj.save(filepath)
+    yield name, filepath, fileobj.filename
```
adamhadani: Do we want to put the yield in a `try: yield ..  except: rmtree(tempdir)` format to try and ensure temporary file is always cleaned up even if client code crashes?

jessemyers: 👍

adamhadani:
LGTM one comment about potentially catching from yield to cleanup temp files

PR #108, opened 2017-03-03
Add qs parameters to retrieve

netama:
Add query-string support for retrieve
Only added to crud and not to relation, but can be added similarly if needed.

### DIFF
--- a/microcosm_flask/conventions/crud.py
+++ b/microcosm_flask/conventions/crud.py
@@ -3,6 +3,7 @@ Conventions for canonical CRUD endpoints.
 
 """
 from inflection import pluralize
+from marshmallow import Schema
 
 from microcosm_flask.conventions.base import Convention
 from microcosm_flask.conventions.encoding import (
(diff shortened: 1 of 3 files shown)


microcosm_flask/conventions/crud.py
```diff
@@ -153,10 +154,14 @@ def configure_retrieve(self, ns, definition):
         :param definition: the endpoint definition
 
         """
+        request_schema = definition.request_schema or Schema()
+
         @self.graph.route(ns.instance_path, Operation.Retrieve, ns)
+        @qs(request_schema)
```
jessemyers: Minor concern: does registering an empty query string schema cause any existing retrieve definitions to change? Or is this a noop?

jessemyers:
LGTM

microcosm_flask/conventions/crud.py
netama: I'm pretty sure it's a noop, I also intentionally added the tests to an existing route and it retrieve seems to return the expected values (both for success and failure). Anything else I should test?

PR #111, opened 2017-03-09
Log a warning when a service returns 500

jessemyers:
Since our services are all much more well-behaved than they once were, it's reasonable to assume that a 500 error is actually a problem that needs investigation. Logs these are warning.

Also, refactor audit logging code path for better clarity.

Now with unit tests!


microcosm_flask/audit.py
```diff
@@ -62,110 +62,177 @@ def wrapper(*args, **kwargs):
...
     Run a request function under audit.
 
     """
     logger = getLogger("audit")
 
+    request_info = RequestInfo(options, func, request_context)
     response = None
 
-    # always include these fields
-    audit_dict = dict(
```
cdm-ium: A welcome refactor.

cdm-ium:
+1

PR #113, opened 2017-03-26
Add CreateCollection operation

dumpstate:


jessemyers:
Looks good, but I'd like one more thing: we can't have both `Create` and `CreateCollection` for the same namespace; can you add a validation in the CRUD convention to raise an exception in that case? (Or perhaps just a unit test assuming that Flask already prevents us from having duplicate endpoint registrations?)

dumpstate:
Unfortunately Flask prevents only duplication on per endpoint (Flask endpoint like `foo.create_collection.v1`) basis, so we can register multiple handlers for same (method, route). I'll add a validation to the CRUDConvention.

jessemyers:
LGTM

PR #117, opened 2017-04-04
Allow custom messages for health checks in /api/health

jessemyers:


afallou:
+1

microcosm_flask/tests/conventions/test_health.py
```diff
@@ -52,7 +52,7 @@ def test_health_check_custom_check():
         "ok": True,
         "checks": {
             "foo": {
-                "message": "ok",
```
cdm-ium: This actually surprises me a little, isn't the default result `ok`?

cdm-ium:
+1

microcosm_flask/tests/conventions/test_health.py
jessemyers: See L45

PR #118, opened 2017-04-11
Initial metrics support.

jessemyers:
Enable timing for routes

### DIFF
--- a/microcosm_flask/namespaces.py
+++ b/microcosm_flask/namespaces.py
@@ -40,6 +40,7 @@ class Namespace(object):
                  controller=None,
                  version=None,
                  enable_basic_auth=False,
+                 enable_metrics=False,
                  identifier_type="uuid"):
         """
         :param subject: the target resource (or resource name) of this namespace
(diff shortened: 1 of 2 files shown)


microcosm_flask/namespaces.py
```diff
@@ -40,6 +40,7 @@ def __init__(self,
                  controller=None,
                  version=None,
                  enable_basic_auth=False,
+                 enable_metrics=False,
```
jessemyers: We can enable metrics by namespace.

microcosm_flask/routing.py
```diff
@@ -21,6 +21,7 @@ def make_path(graph, path):
     enable_audit=True,
     enable_basic_auth=False,
     enable_cors=True,
+    enable_metrics=False,
```
jessemyers: We can enable metrics globally. By default, we do not because it will attempt to resolve the binding key `metrics_timing` and will fail

microcosm_flask/routing.py
```diff
@@ -48,12 +49,17 @@ def route(path, operation, ns):
         :param ns: a `Namespace` instance
         """
         def decorator(func):
+            endpoint = ns.endpoint_for(operation)
```
jessemyers: I'm using the endpoint name as the metric name. This needs some testing.

cdm-ium:
+1

PR #119, opened 2017-04-14
Enable status code counting metrics

jessemyers:
Adds a classifier that looks at status codes.
Also adds unit test coverage.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -11,7 +11,7 @@ general:
 
 dependencies:
   override:
-    - pip install tox tox-pyenv microcosm-metrics
+    - pip install tox tox-pyenv
     - pyenv local 2.7.11 3.6.0  # Should correspond to pre-installed Python versions on CircleCI
 
 test:
--- a/circle.yml
+++ b/circle.yml
@@ -11,7 +11,7 @@ general:
 
 dependencies:
   override:
-    - pip install tox tox-pyenv
+    - pip install tox tox-pyenv microcosm-metrics
     - pyenv local 2.7.11 3.6.0  # Should correspond to pre-installed Python versions on CircleCI
 
 test:
--- a/microcosm_flask/routing.py
+++ b/microcosm_flask/routing.py
@@ -57,15 +57,13 @@ def configure_route_decorator(graph):
         """
         def decorator(func):
             endpoint = ns.endpoint_for(operation)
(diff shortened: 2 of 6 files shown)


circle.yml
```diff
@@ -11,7 +11,7 @@ general:
 
 dependencies:
   override:
-    - pip install tox tox-pyenv
+    - pip install tox tox-pyenv microcosm-metrics
```
jessemyers: Install metrics for unit tests. Otherwise we skip the new tests.

microcosm_flask/routing.py
```diff
@@ -57,16 +57,14 @@ def route(path, operation, ns):
         """
         def decorator(func):
             endpoint = ns.endpoint_for(operation)
+            endpoint_path = make_path(graph, path)
```
jessemyers: Unrelated: I've found myself using this value for debugging here or there, so I figured it was better to calculate it earlier.

microcosm_flask/routing.py
```diff
@@ -77,13 +75,18 @@ def decorator(func):
             # set the opaque component data_func to look at the flask request context
             func = graph.opaque.initialize(graph.request_context)(func)
 
+            if enable_metrics or ns.enable_metrics:
```
jessemyers: Two changes:
 1. I moved the metrics calculation a little later in the chain (audit logs are still last). In theory this makes the measurements a bit more accurate, though in practice we're likely to see big timing differences between the service's internals and the external usage because of network I/O; we'll have to capture that from the next layer up the chain (e.g. Optics)

2. Add a counting decorator so we can track status codes returned by endpoint.

microcosm_flask/tests/test_metrics.py
```diff
@@ -0,0 +1,96 @@
...
+        @self.graph.route(self.ns.collection_path, Operation.Search, self.ns)
+        def foo():
+            return ""
+
+        response = self.client.get("api/v1/foo")
+        assert_that(response.status_code, is_(equal_to(200)))
+
+        self.graph.metrics.histogram.assert_called_with(
+            "example.foo.search.v1",
+            ANY,
```
jessemyers: TIL: `ANY`

KensoDev:
LGTM, I really like this one.

PR #120, opened 2017-04-24
Support marshmallow-polyfield library for polymorphic fields

mitoyarzun:
This library: https://github.com/Bachmann1234/marshmallow-polyfield


jessemyers:
LGTM

PR #122, opened 2017-04-28
Support configuring customized metrics tag mappers on the namespace level

netama:
Tests depend on https://github.com/globality-corp/microcosm-metrics/pull/3.

### DIFF
--- a/microcosm_flask/namespaces.py
+++ b/microcosm_flask/namespaces.py
@@ -33,15 +33,18 @@ class Namespace(object):
 
     """
 
-    def __init__(self,
-                 subject,
-                 object_=None,
-                 path=None,
-                 controller=None,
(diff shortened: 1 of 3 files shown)


microcosm_flask/namespaces.py
```diff
@@ -33,22 +33,27 @@ class Namespace(object):
...
+        self,
+        subject,
+        object_=None,
+        path=None,
+        controller=None,
+        version=None,
+        enable_basic_auth=False,
+        enable_metrics=False,
+        identifier_type="uuid",
+        metrics_tag_mappers=None,
```
jessemyers: As with the comments [here](https://github.com/globality-corp/microcosm-metrics/pull/3), I'd prefer a callable to a mapper.

PR #125, opened 2017-05-09
Add convention to temporarily update logging levels.

jessemyers:
You can now `PATCH` the logging convention endpoint to set logging level for a period
of time. The duration can be capped (default: 5m) and *should* switch back because
of the `ConditionaLLoggingLevel` feature.

### DIFF
--- a/microcosm_flask/conventions/logging_level.py
+++ b/microcosm_flask/conventions/logging_level.py
@@ -3,18 +3,32 @@ Expose/manage logging levels.
 
 """
 from collections import namedtuple
-from logging import PlaceHolder, root
+from logging import getLogger, PlaceHolder, root
+from time import time
 
 from marshmallow import fields, Schema
+from microcosm.api import defaults
+from microcosm_logging.levels import ConditionalLoggingLevel
+from werkzeug.exceptions import UnprocessableEntity
 
 from microcosm_flask.conventions.base import Convention, EndpointDefinition
-from microcosm_flask.conventions.encoding import dump_response_data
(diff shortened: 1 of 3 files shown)


cdm-ium:
+1 hooray

microcosm_flask/conventions/logging_level.py
```diff
@@ -69,28 +84,103 @@ class LoggerSchema(Schema):
     children = fields.List(fields.Nested("LoggerSchema"), required=True)
 
 
-class LoggingLevel(object):
-    pass
+class UpdateConditionalLoggingLevelSchema(Schema):
+    duration = fields.Float(required=False, missing=300.0)
```
cdm-ium: Any real value in being a float here? Wouldn't an int be good enough?

microcosm_flask/conventions/logging_level.py
```diff
@@ -69,28 +84,103 @@ class LoggerSchema(Schema):
...
+    name = fields.String(required=True)
```
cdm-ium: I almost feel like we'll want this to be a list for conveniently setting a batch of loggers in one go.

jessemyers: Let's prove that that's true first.

microcosm_flask/conventions/logging_level.py
jessemyers: There's no reason not to use float since that's the time unit in Python; I chose it because it theoretically lets us write sleep-based tests with sub-second intervals, though I chose to just mock out time.

PR #126, opened 2017-05-10
Use headers to control debug logging instead of timing.

jessemyers:
It turned out that the log level update logic was very impractical because
a real world setting will have many instances spread across multiple processes;
updating the logging levels only takes place in one of these.

A better approach is enable logging per request by a header.

### DIFF
--- a/microcosm_flask/audit.py
+++ b/microcosm_flask/audit.py
@@ -3,8 +3,9 @@ Audit log support for Flask routes.
 
 """
 from collections import namedtuple
+from contextlib import contextmanager
 from functools import wraps
-from logging import getLogger
+from logging import DEBUG, getLogger
 from json import loads
 from traceback import format_exc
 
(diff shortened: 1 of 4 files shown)


microcosm_flask/audit.py
```diff
@@ -45,6 +46,33 @@ def should_skip_logging(func):
     return getattr(func, SKIP_LOGGING, False)
 
 
+@contextmanager
```
jessemyers: This is the new logic. It's simple and powerful.

My main question is: should we use `X-Request-Debug` as the header name so that debugging gets passed through to downstream requests?

 - Advantage: it's useful
 - Disadvantage: passing specific log component names is often useless across different components

One possible solution is to simplify to a boolean value and always set the root logger only.

cdm-ium: I'm fine with using that header name. I think boolean value + root logger makes sense.

microcosm_flask/audit.py
```diff
@@ -45,6 +46,33 @@ def should_skip_logging(func):
...
+    Context manager to conditionally set logging levels.
+
+    Supports setting pre-request, comma-delimited logging levels using the `X-Debug` header.
+    Maps the root logger to "*" for disambiguation.
+
+    """
+    # extra logger names from headers
+    loggers = [
+        "" if logger.strip() == "*" else logger.strip()
+        for logger in request.headers.get("x-debug", "").split(",")
```
cdm-ium: I wonder if we could leverage `graph.opaque` instead of explicitly calling `request.headers` and maybe put all this in `microcosm-logging` (which already depends on microcosm)?

I understand we should try it out first and abstract later but just a thought that occurred to me after thinking we'll want the same thing for the pubsub layer, especially because more of our debug logs will be there.

cdm-ium:
+1

PR #127, opened 2017-05-19
Allow explicit setting of the host to use for service `url_for` resolution

jessemyers:
During some forwarding scenarios (ssh, docker), the inbound host isn't
reachable by the service; this means we can't use `url_for` verbatim

### DIFF
--- a/microcosm_flask/forwarding.py
+++ b/microcosm_flask/forwarding.py
@@ -19,21 +19,36 @@ def use_forwarded_port(graph):
     The URL adapter is used by `url_for` to build a URLs.
 
     """
-    forwarded_port = request.headers.get("X-Forwarded-Port")
-    if not forwarded_port:
-        return None
+    # There must be a better way!
+    context = _request_ctx_stack.top
 
     if _request_ctx_stack is None:
         return None
 
-    # There must be a better way!
-    context = _request_ctx_stack.top
-    context.url_adapter.server_name = "{}:{}".format(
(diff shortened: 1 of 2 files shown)


microcosm_flask/forwarding.py
```diff
@@ -19,21 +19,36 @@ def use_forwarded_port(graph):
...
         return None
 
-    # There must be a better way!
-    context = _request_ctx_stack.top
-    context.url_adapter.server_name = "{}:{}".format(
-        context.url_adapter.server_name.split(":", 1)[0],
-        forwarded_port,
-    )
+    # determine the configured overrides
+    forwarded_host = graph.config.port_forwarding.get("host")
```
jessemyers: I piggy-backed on existing "port forwarding" functionality. It's a poor naming, but it's already used everywhere.

microcosm_flask/tests/test_forwarding.py
```diff
@@ -9,27 +9,52 @@
...
+            port_forwarding=dict(
+                host="service",
+            )
+        )
+        client = self.init(loader)
+
+        response = client.get(
+            "/",
+            headers={
+                "X-Forwarded-Port": "8080",
```
jessemyers: If we configure an explicit host; we use it, ignoring the forwardeing port.

tomgoren:
+1 I think this is a good solution until we have more sophisticated service discovery.

PR #128, opened 2017-05-31
Use different page for links, serialize query string list

pawmarkor:
while working on https://globality.atlassian.net/browse/GLOB-9184 I encountered the following problems:

context: https://github.com/globality-corp/oscar/pull/645/files#diff-4035f8cb0ece7423eb6984cdd98decaaR23

problems:

1. when we use `microcosm_flask.fields.query_string_list.QueryStringList` list items are not deserialized, so when I sent a number e.g. `1` in query string it was deserialized to `1` not to `SystemNoun.SOMETHING`

2. when we want to send a list of enums in query string and this enum is int based e.g. https://github.com/globality-corp/marquez/blob/master/marquez/enums/chatroom_event.py#L9 we want to deserialize the incoming value into enum object but then we got values like `SystemNoun.MESSAGE` instead of `1` in the query string of links in the REST response

3. when we send a list in query string e.g. `a,b,c` we got in the REST response link query string `[a,b,c]` that cannot be parsed back by `QueryStringList`

### DIFF
--- a/microcosm_flask/conventions/crud.py
+++ b/microcosm_flask/conventions/crud.py
@@ -54,9 +54,9 @@ class CRUDConvention(Convention):
         @qs(definition.request_schema)
         @response(paginated_list_schema)
         def search(**path_data):
-            page = self.page_cls.from_query_string(definition.request_schema)
-            result = definition.func(**merge_data(path_data, page.to_dict(func=identity)))
-            response_data, headers = page.to_paginated_list(result, ns, Operation.Search)
+            in_page, out_page = self.page_cls.from_query_string(definition.request_schema)
+            result = definition.func(**merge_data(path_data, in_page.to_dict(func=identity)))
+            response_data, headers = out_page.to_paginated_list(result, ns, Operation.Search)
             return dump_response_data(paginated_list_schema, response_data, headers=headers)
 
         search.__doc__ = "Search the collection of all {}".format(pluralize(ns.subject_name))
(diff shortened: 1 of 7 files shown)


pawmarkor:
@jessemyers @wheremiah please take a look on this

microcosm_flask/fields/query_string_list.py
```diff
@@ -24,6 +24,9 @@ def _deserialize(self, value, attr, obj):
             attribute_elements = [attr_element.split(",") for attr_element in obj.getlist(attr)]
             attribute_params = [param for attr_param in attribute_elements for param in attr_param]
 
-            return attribute_params
+            return super(QueryStringList, self)._deserialize(attribute_params, attr, obj)
```
pawmarkor: related to 1. problem

microcosm_flask/conventions/crud.py
```diff
@@ -54,9 +54,9 @@ def configure_search(self, ns, definition):
         @qs(definition.request_schema)
         @response(paginated_list_schema)
         def search(**path_data):
-            page = self.page_cls.from_query_string(definition.request_schema)
-            result = definition.func(**merge_data(path_data, page.to_dict(func=identity)))
-            response_data, headers = page.to_paginated_list(result, ns, Operation.Search)
+            in_page, out_page = self.page_cls.from_query_string(definition.request_schema)
```
jessemyers: I don't like changing the API on the `Page.from_query_string`. Can you elaborate why you did this?

jessemyers:
I think you might have been better off treating each problem in a separate PR. I understand problem 3 and I see both the solution and a test case that changes as the result (so that is good).

For problems 1 and 2, I'm less sympathetic with the problem (I strongly prefer string-valued enums to integer ones) and I don't follow how the changes fix the problem or see test cases that would convince me that they are solved.

pawmarkor:
@jessemyers you are right, I'm splitting the PR, please take a look on:
https://github.com/globality-corp/microcosm-flask/pull/129

microcosm_flask/conventions/crud.py
pawmarkor: please see my comments in https://github.com/globality-corp/microcosm-flask/pull/130

PR #129, opened 2017-06-01
Deserialize QueryStringList items

pawmarkor:
when we use `microcosm_flask.fields.query_string_list.QueryStringList` list items are not deserialized, please see the added test case

PR #130, opened 2017-06-01
Another page for output qs generation + better serialization for QueryStringList

pawmarkor:
In case we can neglect wrong query strings in search REST response links I don't need this PR

1. when we send a list in query string e.g. `a,b,c` we got in the REST response self link `[a,b,c]` that cannot be parsed back by QueryStringList properly

2. when we want to send a list of enums in query string and this enum is int based e.g. https://github.com/globality-corp/marquez/blob/master/marquez/enums/chatroom_event.py#L9 we want to deserialize the incoming value into enum object but then we got values like `SystemNoun.MESSAGE` instead of `1` in the query string of links in the REST response (showing this require https://github.com/globality-corp/microcosm-flask/pull/129)

### DIFF
--- a/microcosm_flask/conventions/crud.py
+++ b/microcosm_flask/conventions/crud.py
@@ -54,9 +54,9 @@ class CRUDConvention(Convention):
         @qs(definition.request_schema)
         @response(paginated_list_schema)
         def search(**path_data):
-            page = self.page_cls.from_query_string(definition.request_schema)
-            result = definition.func(**merge_data(path_data, page.to_dict(func=identity)))
-            response_data, headers = page.to_paginated_list(result, ns, Operation.Search)
+            in_page, out_page = self.page_cls.from_query_string(definition.request_schema)
+            result = definition.func(**merge_data(path_data, in_page.to_dict(func=identity)))
+            response_data, headers = out_page.to_paginated_list(result, ns, Operation.Search)
             return dump_response_data(paginated_list_schema, response_data, headers=headers)
 
         search.__doc__ = "Search the collection of all {}".format(pluralize(ns.subject_name))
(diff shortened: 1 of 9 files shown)


microcosm_flask/conventions/crud.py
```diff
@@ -54,9 +54,9 @@ def configure_search(self, ns, definition):
         @qs(definition.request_schema)
         @response(paginated_list_schema)
         def search(**path_data):
-            page = self.page_cls.from_query_string(definition.request_schema)
-            result = definition.func(**merge_data(path_data, page.to_dict(func=identity)))
-            response_data, headers = page.to_paginated_list(result, ns, Operation.Search)
```
pawmarkor: Here we create the response page based on the same `page` as we use for actual computing. If the schema contain `QueryStringList` (or sth similar) the deserialized value in the `page` is of type `List` so it got serialized by https://github.com/pawmarkor/microcosm-flask/blob/3aeda476d8c2f0cc1e615a136d778c9a83715d24/microcosm_flask/paging.py#L174 into a list based string i.e. `[a,b,c]` so `QueryStringList` is not used for serialization at all in this case and we got a wrong query string in the response link.

microcosm_flask/paging.py
```diff
@@ -219,7 +221,11 @@ def from_query_string(cls, schema, qs=None):
 
         """
         dct = load_query_string_data(schema, qs)
-        return cls.from_dict(dct)
+        out_data = schema.dump(dct)
```
pawmarkor: Here I use `QueryStringList` for serialization

microcosm_flask/fields/query_string_list.py
```diff
@@ -27,3 +27,6 @@ def _deserialize(self, value, attr, obj):
             return attribute_params
         except ValueError:
             raise ValidationError("Invalid query string list argument")
+
+    def _serialize(self, value, attr, obj):
```
pawmarkor: This is only reasonable if we use `QueryStringList` for serialization of page query string.

pawmarkor:
@jessemyers please take a look

microcosm_flask/conventions/crud.py
jessemyers: A key detail you are missing here: we can (and do!) override `__str__` in enums.

When we build query string arguments for links, [we provide a func](https://github.com/globality-corp/microcosm-flask/blob/develop/microcosm_flask/paging.py#L155), defaulting to `str` to convert values to something representable in our code. We expect user-provided `Enums` to specify an appropriate `__str__` function.

My guess is that you can avoid this PR entirely by adding `__str__` in the right places?

jessemyers:
I still don't really like having page parsing produce two outputs and I'm hopeful there's an easier workaround by defining `__str__` functions. But I might be missing some subtlety?

microcosm_flask/conventions/crud.py
pawmarkor: Yes, `__str__` is good point, thank you, I'm closing this PR

pawmarkor:
To be continued in https://github.com/globality-corp/microcosm-flask/pull/132

PR #131, opened 2017-06-01
Deserialize QueryStringList items

afallou:
Duplicate of https://github.com/globality-corp/microcosm-flask/pull/129 - our CI doesn't handle branches from forks very well.
When we use microcosm_flask.fields.query_string_list.QueryStringList list items are not deserialized, please see the added test case


jessemyers:
LGTM

PR #132, opened 2017-06-02
Use self serializable list in QueryStringList

pawmarkor:
```
>>> from enum import Enum
>>> class A(Enum):
...     A = 1
...     def __str__(self):
...             return self.name
... 
>>> [A.A]
[<A.A: 1>]
>>> str([A.A])
'[<A.A: 1>]'
>>> str(A.A)
'A'
```

### DIFF
--- a/microcosm_flask/fields/query_string_list.py
+++ b/microcosm_flask/fields/query_string_list.py
@@ -27,8 +27,8 @@ class QueryStringList(List):
 
         try:
             attribute_elements = [attr_element.split(",") for attr_element in obj.getlist(attr)]
-            attribute_params = PrintableList(param for attr_param in attribute_elements for param in attr_param)
+            attribute_params = [param for attr_param in attribute_elements for param in attr_param]
 
-            return attribute_params
+            return PrintableList(super(QueryStringList, self)._deserialize(attribute_params, attr, obj))
         except ValueError:
             raise ValidationError("Invalid query string list argument")
--- a/microcosm_flask/fields/query_string_list.py
+++ b/microcosm_flask/fields/query_string_list.py
@@ -5,6 +5,11 @@ A list field field that supports query string parameter parsing.
 from marshmallow.fields import List, ValidationError
 
 
+class SelfSerializableList(list):
+    def __str__(self):
+        return ",".join(str(item) for item in self)
+
+
 class QueryStringList(List):
     def _deserialize(self, value, attr, obj):
         """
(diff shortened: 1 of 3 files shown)


microcosm_flask/fields/query_string_list.py
```diff
@@ -22,7 +27,7 @@ def _deserialize(self, value, attr, obj):
 
         try:
             attribute_elements = [attr_element.split(",") for attr_element in obj.getlist(attr)]
-            attribute_params = [param for attr_param in attribute_elements for param in attr_param]
+            attribute_params = SelfSerializableList(param for attr_param in attribute_elements for param in attr_param)
```
jessemyers: Two thoughts here:
 1. The naming of the list could be better; something like `PrintableList` might be enough; I don't think "Self" is useful or clarifying
 2. Does the output here need to be a list of typed-values (e.g. enums)? Could we avoid the extra class and just use:
```
[str(item) for item in items]
```

microcosm_flask/tests/conventions/fixtures.py
```diff
@@ -95,8 +95,12 @@ def address_retrieve(id, person_id, address_id):
     return ADDRESS_1
 
 
-def address_search(person_id, offset, limit):
-    return [ADDRESS_1], 1, dict(person_id=person_id)
+def address_search(person_id, offset, limit, list_param=None, enum_param=None):
+    if list_param is None or enum_param is None:
+        return [ADDRESS_1], 1, dict(person_id=person_id)
+    return Address(ADDRESS_ID_1,
```
jessemyers: Please indent using:
```
return Address(
    ADDRESS_ID_1,
    ...
)
```

pawmarkor: done

microcosm_flask/fields/query_string_list.py
pawmarkor: 1. done
2. this is deserialization so we want to return a list because we want to have list as internal data type, this deserialization is used for changing query string parameter into internal attribute and we want this list to be able to print nicely back via `str`

jessemyers:
LGTM

pawmarkor:
@jessemyers @afallou I think I'm not able to merge it because of missing CircleCi, could you please help?

pawmarkor:
@jessemyers I'm not able to merge this since it was done in a fork, please accept https://github.com/globality-corp/microcosm-flask/pull/133/

PR #136, opened 2017-08-14
Add format to path parameter

jessemyers:
Ensure that UUIDs are UUIDs.

### DIFF
--- a/microcosm_flask/namespaces.py
+++ b/microcosm_flask/namespaces.py
@@ -37,6 +37,7 @@ class Namespace(object):
                  subject,
                  object_=None,
                  path=None,
+                 prefix=None,
                  controller=None,
                  version=None,
                  enable_basic_auth=False,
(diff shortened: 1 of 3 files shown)


microcosm_flask/namespaces.py
```diff
@@ -37,6 +37,7 @@ def __init__(self,
                  subject,
                  object_=None,
                  path=None,
+                 prefix=None,
```
jessemyers: Unrelated. I'm hoping to remove the `path` argument here.

microcosm_flask/swagger/definitions.py
```diff
@@ -168,17 +168,27 @@ def query_param(name, field, required=False):
...
 
 
-def path_param(name, param_type="string"):
+def path_param(name, ns):
     """
     Build a path parameter definition.
 
     """
-    return swagger.PathParameterSubSchema(**{
+    if ns.identifier_type == "uuid":
```
jessemyers: At the moment, this will only handle string and uuid arguments well, but that's still a win.

PR #137, opened 2017-08-14
Remove several deprecated code paths

jessemyers:
- Stop using "path" for namespaces; use prefix instead
 - Stop trying to configure conventions with custom prefixes
 - Stop trying to auto-generate Namespaces

### DIFF
--- a/microcosm_flask/conventions/build_info.py
+++ b/microcosm_flask/conventions/build_info.py
@@ -42,7 +42,6 @@ class BuildInfoConvention(Convention):
 
 
 @defaults(
-    path_prefix="",
     build_num=None,
     sha1=None,
 )
@@ -52,7 +51,6 @@ def configure_build_info(graph):
 
     """
     ns = Namespace(
-        path=graph.config.build_info_convention.path_prefix,
         subject=BuildInfo,
     )
 
--- a/microcosm_flask/conventions/relation.py
+++ b/microcosm_flask/conventions/relation.py
@@ -18,7 +18,6 @@ from microcosm_flask.conventions.encoding import (
     require_response_data,
 )
 from microcosm_flask.conventions.registry import qs, request, response
-from microcosm_flask.namespaces import Namespace
 from microcosm_flask.operations import Operation
 from microcosm_flask.paging import OffsetLimitPage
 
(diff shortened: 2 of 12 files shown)


microcosm_flask/conventions/build_info.py
```diff
@@ -42,7 +42,6 @@ def build_info():
 
 
 @defaults(
-    path_prefix="",
```
jessemyers: In practice, we don't use the ability to re-prefix the convention-driven endpoints; instead, we use a different `Namespace`.

Removing this "feature" helps clarify the `Namespace` API, which is currently a little confusing.

microcosm_flask/conventions/relation.py
```diff
@@ -205,11 +204,10 @@ def search(**path_data):
         search.__doc__ = "Search for {} relative to a {}".format(pluralize(ns.object_name), ns.subject_name)
 
 
-def configure_relation(graph, ns, mappings, path_prefix=""):
+def configure_relation(graph, ns, mappings):
     """
     Register relation endpoint(s) between two resources.
 
     """
-    ns = Namespace.make(ns, path=path_prefix)
```
jessemyers: Stop trying to generate `Namespace` arguments on-the-fly.

This ability was added to transition to the `Namespace` class, which is now widely used.

microcosm_flask/linking.py
```diff
@@ -87,9 +88,7 @@ def for_(cls, operation, ns, qs=None, type=None, allow_templates=False, **kwargs
         :param kwargs: optional endpoint expansion arguments (e.g. for URI parameters)
         :raises BuildError: if link templating is needed and disallowed
         """
-        # ensure that we actually have a Namespace; many legacy code paths use strings or tuples
-        ns = Namespace.make(ns)
-
+        assert isinstance(ns, Namespace)
```
jessemyers: Catch configuration problems early.

microcosm_flask/namespaces.py
```diff
@@ -32,11 +32,9 @@ class Namespace(object):
     object encapsulates the rest.
 
     """
-
     def __init__(self,
                  subject,
                  object_=None,
-                 path=None,
```
jessemyers: Stop allowing "path" as an argument. It hasn't existed as a real argument for a long time.

PR #141, opened 2017-11-15
Improve audit logging in debug/testing

jessemyers:
Key changes:
 -  Hide request and response bodies over a certain length
 -  Don't log errors as dictionaries
 -  Do log errors with `exc_info`

### DIFF
--- a/microcosm_flask/audit.py
+++ b/microcosm_flask/audit.py
@@ -23,6 +23,10 @@ from microcosm_flask.errors import (
 from microcosm_logging.timing import elapsed_time
 
 
+DEFAULT_INCLUDE_REQUEST_BODY = 400
+DEFAULT_INCLUDE_RESPONSE_BODY = 400
+
+
 AuditOptions = namedtuple("AuditOptions", [
     "include_request_body",
     "include_response_body",
@@ -89,8 +93,8 @@ def audit(func):
     @wraps(func)
     def wrapper(*args, **kwargs):
         options = AuditOptions(
-            include_request_body=True,
-            include_response_body=True,
+            include_request_body=DEFAULT_INCLUDE_REQUEST_BODY,
+            include_response_body=DEFAULT_INCLUDE_RESPONSE_BODY,
             include_path=True,
             include_query_string=True,
         )
@@ -168,7 +172,13 @@ class RequestInfo(object):
     def log(self, logger):
         if self.status_code == 500:
             # something actually went wrong; investigate
-            logger.warning(self.to_dict())
+            dct = self.to_dict()
+
+            if current_app.debug or current_app.testing:
+                message = dct.pop("message")
+                logger.warning(message, extra=dct, exc_info=True)
+            else:
+                logger.warning(dct)
         else:
             # usually log at INFO; a raised exception can be an error or expected behavior (e.g. 404)
             logger.info(self.to_dict())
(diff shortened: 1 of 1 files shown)


microcosm_flask/audit.py
```diff
@@ -89,8 +93,8 @@ def audit(func):
     @wraps(func)
     def wrapper(*args, **kwargs):
         options = AuditOptions(
-            include_request_body=True,
-            include_response_body=True,
+            include_request_body=DEFAULT_INCLUDE_REQUEST_BODY,
```
jessemyers: I am reinterpreting these config parameters as integer size limits (though boolean will still be allowed).

As far as I know, we never configure these today.

The current default is arbitrary.

microcosm_flask/audit.py
```diff
@@ -168,7 +172,13 @@ def to_dict(self):
     def log(self, logger):
         if self.status_code == 500:
             # something actually went wrong; investigate
-            logger.warning(self.to_dict())
+            dct = self.to_dict()
+
+            if current_app.debug or current_app.testing:
+                message = dct.pop("message")
```
jessemyers: In debug/test, log errors so that:
 - We have a clear message
 - We have a stack trace
 - We pass everything else as extra

microcosm_flask/audit.py
```diff
@@ -342,7 +367,7 @@ def login(username, password):
             ...
     """
     include_request_body = graph.config.audit.include_request_body
-    include_response_body = graph.config.audit.include_response_body
+    include_response_body = int(graph.config.audit.include_response_body)
```
cdm-ium: So this probably wouldn't have really worked beore if we would receive it as a string from the ENV or elsewhere?

jessemyers: yes. and I need to add the missing `int` above.

microcosm_flask/audit.py
```diff
@@ -182,6 +192,14 @@ def capture_request(self):
             # only capture request body if requested
             return
 
+        if all((
+            request.content_length,
+            self.options.include_request_body is not True,
+            request.content_length >= self.options.include_request_body,
+        )):
+            # don't capture request body if it's too large
```
pelegdvir: If its json - you can capture the root keys instead

microcosm_flask/audit.py
```diff
@@ -205,6 +223,13 @@ def capture_response(self, response):
             # only capture request body if there is one
             return
 
+        if all((
+            self.options.include_response_body is not True,
+            len(body) >= self.options.include_response_body,
```
cdm-ium: Just being paranoid, but will this comparison fire in some weird way if `self.options.include_response_body == False`?

microcosm_flask/audit.py
jessemyers: If you have a specific algorithm in mind, I'm all ears. I started trying to think about how cleanly truncate a nested JSON document... and represent the fact that it was truncated (e.g. with `...`) and still keep it as valid JSON... and went with the simple thing.

microcosm_flask/audit.py
jessemyers: Nope:

```
>>> 1 >= False
True
>>> 0 >= False
True
>>>
```

If it's false, this evaluates to `True` and we skip, which is the meaning of `incldue_rsponse_body=False` in this case.

PR #143, opened 2017-12-04
Add CSV convention with search route

afallou:
- Add CSV convention
- Update `encoding` to allow dumping a response to CSV
- When dumping response, take into account schema's `csv_column_order` property if present
- Pass `csv_column_order` through in `PaginatedListSchema` class

**Update**
- Offload building of the flask `Response` to a `formatter`
- Add formatters for JSON and CSV
- Default to JSON formatter
- Remove CSV convention and instead add content negotiation logic to `CRUDConvention`


afallou:
The test I added succeeds on its own, but fails when run with the other ones; probably because of some leftover state. Trying to figure out why.

microcosm_flask/conventions/csv_convention.py
```diff
@@ -0,0 +1,84 @@
...
+
+    :param mappings: a dictionary from operations to tuple, where each tuple contains
+                     the target function and zero or more marshmallow schemas according
+                     to the signature of the "register_<foo>_endpoint" functions
+
+    Example mapping:
+
+        {
+            Operation.Create: (create_foo, NewFooSchema(), FooSchema()),
+            Operation.Delete: (delete_foo,),
```
jessemyers: What does it mean to delete a CSV? I assume create is a future goal?

microcosm_flask/conventions/encoding.py
```diff
@@ -113,7 +116,7 @@ def remove_null_values(data):
     return data
 
 
-def dump_response_data(response_schema, response_data, status_code=200, headers=None):
+def dump_response_data(response_schema, response_data, status_code=200, headers=None, response_format=None):
```
jessemyers: I think you'll make this cleaner if you abstract the format in to a `formatter` "dumps" function and have the CSV convention pass the formatter instead of text string. You can have a default formatter passed as the kwargs default or you can default to JSON if None.

microcosm_flask/conventions/encoding.py
```diff
@@ -126,20 +129,38 @@ def dump_response_data(response_schema, response_data, status_code=200, headers=
...
+def make_response(response_data, response_schema=None, status_code=200, headers=None, response_format=None):
     if request.headers.get("X-Response-Skip-Null"):
         # swagger does not currently support null values; remove these conditionally
         response_data = remove_null_values(response_data)
 
     headers = headers or {}
-    if "Content-Type" not in headers:
-        # Specify JSON as the response content type by default
-        headers["Content-Type"] = "application/json"
+    if response_format == "csv":
```
jessemyers: Per above:

`response, headers = formatter(...)`

microcosm_flask/conventions/encoding.py
```diff
@@ -170,3 +191,36 @@ def require_response_data(response_data):
     if not response_data:
         raise NotFound
     return response_data
+
+
+def csvify(response_data, response_schema):
```
jessemyers: It would be wise to organize csv code into another module, especially if you break up formatting into a base/json/csv scheme.

microcosm_flask/conventions/csv_convention.py
afallou: Leftover from a copy-paste; I don't think we really need `create` or `delete` (especially since delete wouldn't differ between CSV and json)

microcosm_flask/conventions/encoding.py
```diff
@@ -170,3 +191,36 @@ def require_response_data(response_data):
...
+    if column_order is None:
+        # We should still be able to return a CSV even if no column order has been specified
+        column_names = response_fields
+    else:
+        column_names = response_schema.csv_column_order
+        # The column order be only partially specified
+        column_names.extend([field_name for field_name in response_fields if field_name not in column_names])
+
+    output = StringIO()
+    csv_writer = writer(output, quoting=QUOTE_MINIMAL)
```
jessemyers: I recommend testing this with a large data set. Memory usage could go up if we try to dump the entire result set into a string buffer. Might be ok, might not (since our containers usually have a memory cap)

microcosm_flask/conventions/encoding.py
```diff
@@ -170,3 +191,36 @@ def require_response_data(response_data):
...
     return response_data
+
+
+def csvify(response_data, response_schema):
+    """
+    Make Flask `response` object, with data returned as a generator for the CSV content
+    The CSV is built from JSON-like object (Python `dict` or list of `dicts`)
+
+    """
+    # TODO: determine column ordering with response schema
```
jessemyers: Does a `DictWriter` help?

microcosm_flask/runserver.py
```diff
@@ -12,6 +12,7 @@ def parse_args(graph):
     default_enable_profiling = graph.config.flask.enable_profiling
 
     parser = ArgumentParser()
+    parser.add_argument("-s", default="127.0.0.1")
```
jessemyers: What's this for?

microcosm_flask/tests/conventions/test_csv.py
```diff
@@ -0,0 +1,106 @@
...
+    Operation.Search: (person_search, OffsetLimitPageSchema(), PersonCSVSchema()),
+}
+
+
+ADDRESS_MAPPINGS = {
+    Operation.Search: (address_search, SearchAddressPageSchema(), AddressCSVSchema()),
+}
+
+
+class TestCSV(object):
```
jessemyers: object not required

microcosm_flask/runserver.py
afallou: leftover from manual testing - will remove

microcosm_flask/tests/conventions/test_csv.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def assert_csv_response(self, response, status_code, expected_lines=None):
+        # always validate status code
+        assert_that(response.status_code, is_(equal_to(status_code)))
+
+        # expect JSON data except on 204
+        if status_code == 204:
+            response_lines = None
+        else:
+            response_lines = [row for row in reader(StringIO(response.data))]
```
jessemyers: This doesn't feel like the optimal way to compare CSVs.

microcosm_flask/tests/conventions/test_csv.py
afallou: I thought we wanted compatibility with Python 2 and 3?

jessemyers: Fair.

wheremiah:
I like the idea of being able to return the data in different formats.
I would have expected this to be done using the `Accept: text/csv` header using an existing end point, rather than linking the type of data returned to the endpoint.

jessemyers:
@wheremiah Interesting idea.

I think this would look like:
 - An abstraction around formatting as described in the current PR comments
 - A CSV implementation of the formatter
 - Content negotiation logic for CRUD endpoints based on Accept header
 - A way to define the default formatter (in the `Namespace`? ) as needed

The key questions are whether we'd need to override the formatters functions or mappings in some way and what new graph components we'd need to make that sane.

afallou:
@wheremiah good point; I guess the way I built it corresponds to the usage we've had so far, with separate endpoints for specific CSV reports. But then again, we also built separate routes because we couldn't use `microcosm-flask` route definitions to return csv.

afallou:
@wheremiah @jessemyers I implemented the changes you proposed.

microcosm_flask/conventions/base.py
```diff
@@ -42,6 +45,10 @@ def response_schema(self):
     def header_func(self):
         return self[3] or identity
 
+    @property
+    def response_formats(self):
+        return self[4]
```
jessemyers: Do you want an `or []` here?

microcosm_flask/conventions/encoding.py
```diff
@@ -170,3 +178,30 @@ def require_response_data(response_data):
...
+
+    If the 'Accept' header doesn't exactly match a format we can handle, we return JSON
+
+    """
+    if allowed_response_formats is None:
+        allowed_response_formats = [ResponseFormats.JSON]
+
+    request_accept_content_type = request.headers.get("Accept")
+
+    if request_accept_content_type is not None:
```
jessemyers: Invert and return early?

microcosm_flask/enums.py
```diff
@@ -0,0 +1,23 @@
...
+    JSON_CONTENT_TYPE,
+    CSV_CONTENT_TYPE,
+)
+
+ResponseFormatSpec = namedtuple("ResponseFormatSpec", ["content_type", "formatter"])
+
+
+@unique
+class ResponseFormats(Enum):
+    CSV = ResponseFormatSpec(
```
jessemyers: Weird alignment

microcosm_flask/conventions/crud.py
```diff
@@ -60,7 +68,13 @@ def search(**path_data):
             result = definition.func(**merge_data(path_data, page.to_dict(func=identity)))
             response_data, headers = page.to_paginated_list(result, ns, Operation.Search)
             definition.header_func(headers)
-            return dump_response_data(paginated_list_schema, response_data, headers=headers)
+            response_format = self.negotiate_response_content(definition.response_formats)
```
jessemyers: You probably also need to instrument the relation convention routes...

There's probably also a refactoring we could do to make the endpoint smarter so we do something more like: `endpoint.dump(...)` and then could make this kind of change in one place instead of many. I assume this is out of scope.

microcosm_flask/conventions/base.py
afallou: `or []`. I wanted to avoid having the "default to JSON" behavior expressed in this class, since it's just a thin wrapper around a tuple, and should be agnostic to our conventions.

PR #144, opened 2017-12-20
Handle form body content during multipart upload

jessemyers:
Handling JSON content would be nice too, but isn't easy to test using `flask/werkzeug`
since their [test clients](https://github.com/pallets/werkzeug/blob/master/werkzeug/test.py#L349) assume form-encoded for other content once files are present.

### DIFF
--- a/microcosm_flask/tests/conventions/test_upload.py
+++ b/microcosm_flask/tests/conventions/test_upload.py
@@ -6,10 +6,12 @@ from uuid import uuid4
 
 from hamcrest import (
     all_of,
+    anything,
     assert_that,
     contains,
     equal_to,
     has_entry,
+    has_entries,
     has_item,
     has_key,
     is_,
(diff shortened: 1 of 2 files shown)


microcosm_flask/tests/conventions/test_upload.py
```diff
@@ -167,3 +195,26 @@ def test_upload_for(self):
...
+            data=dict(
+                file=(BytesIO(b("Hello World\n")), "hello.txt"),
+                extra="special",
+            ),
+        )
+        assert_that(response.status_code, is_(equal_to(204)))
+        assert_that(self.controller.calls, contains(
+            has_entries(
+                files=contains(contains("file", anything(), "hello.txt")),
+                extra="special",
```
jessemyers: This is the main thing.

Previously, we'd pass non-file parameters in the query string; now we can easily pull them from the form body too.

wheremiah: Nice

PR #147, opened 2018-01-08
Change condition to find config function in base Convention

afallou:
The condition was for the passed `operation` to be an instance of `Operation`. However, `Operation` is an enum defined in `microcosm-flask`; since Enums cannot be extended, it makes it impossible for a user of `microcosm-flask` to define their own `Operation`-like enum and use its members in a endpoint mapping.

### DIFF
--- a/microcosm_flask/conventions/base.py
+++ b/microcosm_flask/conventions/base.py
@@ -2,10 +2,11 @@
 Convention base class.
 
 """
+from enum import Enum
+
 from werkzeug.exceptions import NotAcceptable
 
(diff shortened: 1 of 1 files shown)


jessemyers:
I'm not convinced we want extensible `Operations`. The point of a strongly convention-driven approach is that we have to have a discussion/debate in a PR in order to add a new operation; different services can't just make up their own conventions.

Note that we do have the general `Command` and `Query` operations as a fail-safe if the lack of an operation would have blocked progress.

afallou:
:+1:

PR #149, opened 2018-01-16
Calculate identifier type for swagger definition by parsing rule

saivickna:


saivickna:
Will add unit tests after review

PR #150, opened 2018-01-23
Add anonymize and replace support for cloning

netama:


jessemyers:
Are unit tests coming?

PR #151, opened 2018-01-30
Revamp formatters

jessemyers:
Changes:

 - Clearer/simpler base class implementation
 - Less code in subclasses
 - Tests

### DIFF
--- a/microcosm_flask/formatting/base.py
+++ b/microcosm_flask/formatting/base.py
@@ -1,17 +1,39 @@
+"""
+Response formatting base class.
+
+"""
 from abc import ABCMeta, abstractmethod
+
+from flask import Response
+from werkzeug.utils import get_content_type
 
 
 class BaseFormatter(metaclass=ABCMeta):
+
     def __init__(self, response_schema=None):
         # Formatting could need the response schema
         # e.g. to specify column ordering in CSV response
         self.response_schema = response_schema
 
-    @abstractmethod
-    def make_response(self, response_data, headers):
-        """
-        Build network response object from response data `dict`
-        The user can also pass in additional headers
+    def __call__(self, response_data, headers=None, **kwargs):
+        response = self.build_response(response_data)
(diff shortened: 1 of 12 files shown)


microcosm_flask/formatting/base.py
```diff
@@ -1,17 +1,34 @@
...
-        # Formatting could need the response schema
-        # e.g. to specify column ordering in CSV response
-        self.response_schema = response_schema
 
-    @abstractmethod
-    def make_response(self, response_data, headers):
-        """
-        Build network response object from response data `dict`
-        The user can also pass in additional headers
+    def __call__(self, response_data, headers=None, **kwargs):
```
jessemyers: Refactoring here to delegate to `build_response` and `build_headers`

microcosm_flask/tests/formatting/test_json_formatter.py
```diff
@@ -0,0 +1,28 @@
...
+
+def test_make_response():
+    graph = create_object_graph(name="example", testing=True)
+    formatter = JSONFormatter()
+
+    with graph.app.test_request_context():
+        response = formatter(dict(foo="bar"))
+
+    assert_that(response.data, is_(equal_to(b'{\n  "foo": "bar"\n}\n')))
+    assert_that(response.content_type, is_(equal_to("application/json")))
```
jessemyers: Note that the existing behavior in `jsonify` does not add an explicit `charset=utf-8` encoding to the content type. This seems weird, but isn't new, so I'm not changing it here.

microcosm_flask/formatting/base.py
```diff
@@ -1,17 +1,34 @@
...
+        response = self.build_response(response_data)
+        headers = self.build_headers(headers=headers or {}, **kwargs)
+        response.headers.extend(headers)
```
netama: any point worrying about duplicate headers?

jessemyers: Not really.

PR #153, opened 2018-02-08
Let services expose their config via api

cdm-ium:
Why? One major impediment we've had to users ramping up in our environment is figuring out the correct config for a given service. One thing we discussed was just hosting valid config for services somewhere with some amount of security. Here's one way we could do it.

- config is protected by basic auth
- sharing config is *disabled* by default so we can opt in only for testing environments
- we could have different auth per service or per environment. My vote is per environment so that developers can set up config once for being able to pull config from dev and then set up is very minimal.
- services that wish to use this need to add the `config_convention`

Usage is very simple:
```
http -a default:secret https://dev-service/api/config > dev-service.conf
DEV_SERVICE__SETTINGS=/Users/dino/globality/dev-service/dev-service.conf runserver
```

### DIFF
--- a/microcosm_flask/conventions/config.py
+++ b/microcosm_flask/conventions/config.py
@@ -8,7 +8,7 @@ locally with realistic config.
 from distutils.util import strtobool
 from json import dumps, loads
 
-from microcosm.api import defaults
+from microcosm.api import defaults, typed, required
 from microcosm.loaders.compose import PartitioningLoader
 from microcosm_flask.audit import skip_logging
 from microcosm_flask.conventions.base import Convention
@@ -66,7 +66,7 @@ class ConfigDiscoveryConvention(Convention):
 
 
 @defaults(
-    enabled="False",
+    enabled=typed(bool, default_value=False),
 )
 def configure_config(graph):
     """
--- a/microcosm_flask/conventions/config.py
+++ b/microcosm_flask/conventions/config.py
@@ -57,7 +57,6 @@ class ConfigDiscoveryConvention(Convention):
             return
 
         @self.add_route(ns.singleton_path, Operation.Retrieve, ns)
-        @self.graph.basic_auth.required
         @skip_logging
         def current_config_discovery():
             response_data = self.config_discovery.to_dict()
@@ -75,6 +74,7 @@ def configure_config(graph):
     """
     ns = Namespace(
         subject=Config,
+        enable_basic_auth=True,
     )
 
     convention = ConfigDiscoveryConvention(
--- a/microcosm_flask/conventions/config.py
+++ b/microcosm_flask/conventions/config.py
@@ -9,12 +9,9 @@ from distutils.util import strtobool
 from json import dumps, loads
 
 from microcosm.api import defaults
-from microcosm.object_graph import config_report
 from microcosm_flask.audit import skip_logging
 from microcosm_flask.conventions.base import Convention
-from microcosm_flask.conventions.build_info import BuildInfo
 from microcosm_flask.conventions.encoding import make_response
-from microcosm_flask.errors import extract_error_message
 from microcosm_flask.namespaces import Namespace
 from microcosm_flask.operations import Operation
 
@@ -77,7 +74,6 @@ def configure_config(graph):
     ns = Namespace(
         subject=Config,
     )
-    print(ns.singleton_path)
 
     convention = ConfigDiscoveryConvention(
         graph,
--- a/microcosm_flask/conventions/config.py
+++ b/microcosm_flask/conventions/config.py
@@ -5,10 +5,9 @@ Reports service config protected by basic auth for securely running services
 locally with realistic config.
 
 """
-from distutils.util import strtobool
 from json import dumps, loads
 
-from microcosm.api import defaults, typed, required
+from microcosm.api import defaults, typed
 from microcosm.loaders.compose import PartitioningLoader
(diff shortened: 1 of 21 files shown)


microcosm_flask/conventions/config.py
```diff
@@ -0,0 +1,83 @@
...
+
+    """
+    def __init__(self, graph, include_build_info=True):
+        self.graph = graph
+        self.name = graph.metadata.name
+        self.config = dict(
+            self.graph.config
+        )
+
+    def to_dict(self):
```
cdm-ium: This is the most questionable part. We could probably inject a custom json encoder in a different part of the stack, but not sure we even like this direction in general so just putting this idea out there with something simple.

microcosm_flask/runserver.py
```diff
@@ -15,6 +15,7 @@ def parse_args(graph):
     parser.add_argument("--host", default="127.0.0.1")
     parser.add_argument("--port", type=int, default=default_port)
     parser.add_argument("--with-profiling", action="store_true", default=default_enable_profiling)
+    parser.add_argument("--debug", action="store_true", default=False)
```
cdm-ium: I started with using `graph.metadata` to toggle flags in the microcosm graph object itself, but it didn't get there. I can remove this, but setting `metadata.debug` might be useful for other features.....

microcosm_flask/conventions/config.py
```diff
@@ -0,0 +1,83 @@
+"""
+Config sharing convention.
+
+Reports service config protected by basic auth for securely running services
+locally with realistic config.
+
+"""
+from distutils.util import strtobool
+from microcosm.api import defaults
```
jessemyers: not a built-in

microcosm_flask/conventions/config.py
```diff
@@ -0,0 +1,83 @@
...
+        self.config = dict(
+            self.graph.config
+        )
+
+    def to_dict(self):
+        """
+        Encode the name, the status of all checks, and the current overall status.
+
+        """
+        def remove_nulls(dct):
```
jessemyers: How come? In theory null values are meaningful.

microcosm_flask/conventions/config.py
jessemyers: I think we need to do some work in the config layer to make this safe (e.g. labeling secrets). We have a few other ideas around config improvements (e.g. validation of required values and auto-typing). We're probably do for a design session there.

wheremiah:
In general I really like this idea.  If we can get the security work out I think this would be very helpful.

microcosm_flask/conventions/config.py
cdm-ium: Well the real problem is that things that were not json in the original dump are converted to `null` before passed off to flask so `null` might mean "some object that was automatically loaded in". Not cleaning out the nulls would mean overridden config loaded within the code--like when we configure the PubsubEnvelope for a daemon and default it to the SQSEnvelope.

cdm-ium:
Reopenning now that we've refactored config in `microcosm`.

cdm-ium:
Tests will fail without microcosm updates in https://github.com/globality-corp/microcosm/pull/27/files.

microcosm_flask/conventions/config.py
```diff
@@ -0,0 +1,86 @@
...
+        if not isinstance(self.graph.loader, PartitioningLoader):
+            return dict(msg="Config sharing disabled for non-partioned loader")
+
+        if not hasattr(self.graph.loader, "secrets"):
+            return dict(msg="Config sharing disabled if no secrets are labelled")
+
+        def remove_nulls(dct):
+            return {key: value for key, value in dct.items() if value is not None}
+
+        return loads(
```
cdm-ium: I could do more work to include a sanitized version of secrets, e.g. 
```
{
config = ...,
secrets=...,
}
```
But as an unintended feature, the defaults for the secrets will still get picked up from the factories so it should be easy enough to edit in the valid values as needed.

microcosm_flask/conventions/config.py
```diff
@@ -0,0 +1,86 @@
...
+        @self.add_route(ns.singleton_path, Operation.Retrieve, ns)
+        @self.graph.basic_auth.required
+        @skip_logging
+        def current_config_discovery():
+            response_data = self.config_discovery.to_dict()
+            return make_response(response_data, status_code=200)
+
+
+@defaults(
+    enabled="False",
```
jessemyers: You should use `typed(boolean, default_value=False)`

microcosm_flask/conventions/config.py
```diff
@@ -0,0 +1,85 @@
...
+        super(ConfigDiscoveryConvention, self).__init__(graph)
+        self.config_discovery = Config(graph)
+        self.enabled = enabled
+
+    def configure_retrieve(self, ns, definition):
+        if not self.enabled:
+            return
+
+        @self.add_route(ns.singleton_path, Operation.Retrieve, ns)
+        @self.graph.basic_auth.required
```
jessemyers: It would probably be better to use the [enable_basic_auth](https://github.com/globality-corp/microcosm-flask/blob/develop/microcosm_flask/namespaces.py#L42) attribute of the `Namespace` and the default routing mechanisms.

Since you control the `Namespace` construction on L76, that means you have the ability to set a reasonable default for basic auth without making alternate configurations impossible.

PR #154, opened 2018-02-13
Add ability to return CSV formatted data without specifying column names

saivickna:
Why?

We want the ability to return unstructured CSV data

### DIFF
--- a/microcosm_flask/formatting/csv_formatter.py
+++ b/microcosm_flask/formatting/csv_formatter.py
@@ -23,6 +23,21 @@ class CSVFormatter(BaseFormatter):
 
         return headers
 
+    def get_column_names(self, list_response_data):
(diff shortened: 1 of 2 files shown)


microcosm_flask/formatting/csv_formatter.py
```diff
@@ -34,23 +34,31 @@ def format(self, response_data):
...
-
-            # The column order be only partially specified
-            column_names.extend([field_name for field_name in response_fields if field_name not in column_names])
+        write_column_names = type(list_response_data[0]) not in (tuple, list)
 
         output = StringIO()
         csv_writer = writer(output, quoting=QUOTE_MINIMAL)
-        csv_writer.writerow(column_names)
+
+        if write_column_names:
```
jessemyers: I would make L43-L55 a separate function.

PR #155, opened 2018-02-22
Remove basic auth and enabled flag.

cdm-ium:
Why? The enabled flag is redundant with deciding to include the component in the
first place, we decided not to share any secret information so basic auth isn't
necessary.

### DIFF
--- a/microcosm_flask/conventions/config.py
+++ b/microcosm_flask/conventions/config.py
@@ -7,7 +7,6 @@ locally with realistic config.
 """
 from json import dumps, loads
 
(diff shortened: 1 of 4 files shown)


microcosm_flask/conventions/config.py
```diff
@@ -7,7 +7,6 @@
 """
 from json import dumps, loads
 
-from microcosm.api import defaults, typed
```
cdm-ium: Some cleanup after discussions with Jesse.

PR #156, opened 2018-03-03
Allow headers to leverage response data

jessemyers:
Enables business logic to generate values used in headers.

### DIFF
--- a/microcosm_flask/conventions/base.py
+++ b/microcosm_flask/conventions/base.py
@@ -23,9 +23,24 @@ class EndpointDefinition(tuple):
     """
     def __new__(cls, func=None, request_schema=None, response_schema=None, header_func=None, response_formats=None):
         """
+        Define an API endpoint.
+
+        Defines the behavior of an API endpoint in conjunction with a `Namespace` and an `Operation`.
+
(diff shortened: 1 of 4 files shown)


microcosm_flask/conventions/base.py
```diff
@@ -23,9 +23,24 @@ class EndpointDefinition(tuple):
...
+
+        The callable `func` should accept `**kwargs` and return a marshmallow-compatible object or dictionary.
+
+        The callable `header_func` (if any) should accept a `headers` dictionary and the return value from the
+        callable `func`.
+
         :param func: a function to process request data and return response data
         :param request_schema: a marshmallow schema to decode/validate request data
         :param response_schema: a marshmallow schema to encode response data
+        :param header_func: a header-modifying function
```
moshez: That's a backwards incompatible change, right? It should at least be marked more prominently as such.

An alternative is to accept a new `header_func_from_response` and deprecate `header_func`, but that might be a big heavy-weight.

jessemyers: Meh. I checked for existing usage in our code base and didn't find any. Supporting hypothetical external developer regressions is not a priority until we have evidence that such developers exist.

PR #158, opened 2018-06-15
Add landing page convention

saivickna:
Why?

Useful for showing links to relevant configs and provide information about a service. Works in tandem with health check convention and config convention.
Can specify a description or github link by by binding 'description' or 'github' to the microcosm services config

### DIFF
--- a/microcosm_flask/conventions/landing.py
+++ b/microcosm_flask/conventions/landing.py
@@ -1,0 +1,22 @@
+"""
+Landing Page convention.
+
+"""
+from microcosm_flask.templates.landing import template
+from jinja2 import Template
+
+
+def configure_landing(graph):
+
+    @graph.flask.route("/")
+    def render_landing_page():
+        """
+        Render landing page
+
+        """
+        return Template(template).render(
+            service_name=graph.metadata.name,
+            swagger_version=graph.config.swagger_convention.version,
+            description=graph.config.get('description', None),
+            github=graph.config.get('github', None),
+        )
--- a/microcosm_flask/conventions/landing.py
+++ b/microcosm_flask/conventions/landing.py
@@ -2,11 +2,63 @@
 Landing Page convention.
 
 """
+from microcosm_flask.conventions.registry import iter_endpoints
 from microcosm_flask.templates.landing import template
 from jinja2 import Template
+from os.path import join
+from pkg_resources import working_set, Requirement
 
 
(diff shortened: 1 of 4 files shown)


microcosm_flask/conventions/landing.py
```diff
@@ -0,0 +1,22 @@
...
+    @graph.flask.route("/")
+    def render_landing_page():
+        """
+        Render landing page
+
+        """
+        return Template(template).render(
+            service_name=graph.metadata.name,
+            swagger_version=graph.config.swagger_convention.version,
+            description=graph.config.get('description', None),
```
jessemyers: This is a good start, but I'd love to make this not require configuration.

One possible direction:
 - See if we can programmatically inspect any of the data set in `setup.py` (e.g. `long_description`)
 - Write something to populate that field in `setup.py` from a `README.md` (and move developer information into a different file, e.g `CONTRIBUTING.md` or `DEVELOPING.md`)

microcosm_flask/conventions/landing.py
```diff
@@ -0,0 +1,22 @@
...
+
+    @graph.flask.route("/")
+    def render_landing_page():
+        """
+        Render landing page
+
+        """
+        return Template(template).render(
+            service_name=graph.metadata.name,
+            swagger_version=graph.config.swagger_convention.version,
```
jessemyers: What if there are multiple versions at once?

We've been discouraging the use of `graph.config.swagger_convention.version` in many/most of our services for this reason.

microcosm_flask/conventions/landing.py
```diff
@@ -0,0 +1,22 @@
...
+            description=graph.config.get('description', None),
+            github=graph.config.get('github', None),
```
jessemyers: As above, probably available in `setup.py`, might be accessible programmatically.

microcosm_flask/conventions/landing.py
```diff
@@ -0,0 +1,79 @@
...
+        """
+        properties = get_properties()
+        description = get_description(properties)
+        swagger_versions = get_swagger_versions()
+
+        return Template(template).render(
+            service_name=graph.metadata.name,
+            swagger_versions=swagger_versions,
+            description=description,
+            github=properties.get("Home-page"),
```
pelegdvir: Use Home instead of github?

PR #178, opened 2018-07-10
Add links section to landing page

sethisernhagen:
- add a section to landing page for listing links specified in config

Why?

Could be used to display links to loggly, grafana, github.

### DIFF
--- a/microcosm_flask/conventions/landing.py
+++ b/microcosm_flask/conventions/landing.py
@@ -28,20 +28,6 @@ def configure_landing(graph):   # noqa: C901
         except DistributionNotFound:
             return None
 
-    def get_description(properties):
-        """
-        Calculate the description based on the package properties
-
-        """
-        if not properties:
-            return None
-
-        return '. '.join([
-            field
-            for field in [properties.description, properties.long_description]
-            if field is not None
-        ])
-
     def get_swagger_versions():
         """
         Finds all swagger conventions that are bound to the graph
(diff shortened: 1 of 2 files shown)


microcosm_flask/templates/landing.py
```diff
@@ -89,6 +84,18 @@
                     <pre id=config>{{ config }}</pre>
                 </div>
             </div>
+            {%-if links -%}
```
jessemyers: nit: `{%- if`

microcosm_flask/conventions/landing.py
```diff
@@ -83,18 +69,18 @@ def render_landing_page():
...
         env = get_env_file_commands(config, graph.metadata.name)
 
         return Template(template).render(
             config=pretty_dict(config),
-            description=description,
+            description=properties.description,
             env=env,
             health=pretty_dict(health),
             homepage=getattr(properties, 'url', None),
+            links=graph.config.landing_convention.get("links", []),
```
jessemyers: We're starting to pull input data from a wide variety of places without any obvious consistency. Do you think we're ready for an abstraction layer (e.g. a controller) and object model between this route and the template?

sethisernhagen: That would enable us to retrieve these links from an endpoint instead of getting it directly from config?

Doing that would enable us to have logic for generating the url? Ex. a link to a grafana dashboard

Making the coupling really loose makes sense to me now but that means that all url logic will need to be in ansible.

I don't know the answer.

PR #182, opened 2018-08-23
Support oneOf for custom fields

jessemyers:


microcosm_flask/swagger/schema.py
```diff
@@ -82,7 +82,17 @@ def resolve_tagged_field(field):
     """
     field_type = getattr(field, SWAGGER_TYPE)
     field_format = getattr(field, SWAGGER_FORMAT, None)
-    if field_format:
+
+    if isinstance(field_type, list):
```
jessemyers: For custom fields, if the swagger type is a list, assume we want one of.

In this case, we'll ignore the field format. We could extend this to support a list of formats and run a zip over them, but that's not needed yet.

PR #191, opened 2018-09-26
GLOB-25359: Move the endpoint to tags rather than the name.

moshez:
Why?
We want to know that the metrics for different endpoints are "commensurate".
For example, we can add all the 200 counts to see how many total 200 counts we have.
This allows us whatever server pulls from the statsd server.

### DIFF
--- a/microcosm_flask/routing.py
+++ b/microcosm_flask/routing.py
@@ -77,10 +77,13 @@ def configure_route_decorator(graph):
 
             if enable_metrics or ns.enable_metrics:
                 from microcosm_flask.metrics import StatusCodeClassifier
-                func = graph.metrics_counting("microcosm_flask",
-                                              tags=["endpoint:" + endpoint],
-                                              classifier_cls=StatusCodeClassifier)(func)
-                func = graph.metrics_timing("microcosm_flask", tags=["endpoint:" + endpoint])(func)
+                tags = [f"endpoint:{endpoint}", "backend_type:microcosm_flask"]
+                func = graph.metrics_counting(
+                    "route",
+                    tags=tags,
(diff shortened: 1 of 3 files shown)


microcosm_flask/routing.py
```diff
@@ -77,8 +77,10 @@ def decorator(func):
 
             if enable_metrics or ns.enable_metrics:
                 from microcosm_flask.metrics import StatusCodeClassifier
-                func = graph.metrics_counting(endpoint, classifier_cls=StatusCodeClassifier)(func)
-                func = graph.metrics_timing(endpoint)(func)
+                func = graph.metrics_counting("microcosm_flask",
```
jessemyers: Preferred indentation style is:

```
func = blahblah(
    this=that,
    foo=bar,
)(func)
```

microcosm_flask/routing.py
```diff
@@ -77,8 +77,10 @@ def decorator(func):
...
+                                              tags=["endpoint:" + endpoint],
+                                              classifier_cls=StatusCodeClassifier)(func)
+                func = graph.metrics_timing("microcosm_flask", tags=["endpoint:" + endpoint])(func)
```
jessemyers: You should a) use f-strings instead of `+` and b) assign `tags=[f"endpoint:{endpoint}"]` so you don't need to declare this twice.

microcosm_flask/routing.py
moshez: Feedback: is `microcosm_flask` a good name for the metric? Would we want a `microcosm_flask_route`? Something else?

moshez: Summarizing slack discussion: `route` for name, `backend_type=microcosm_flask` tag.

PR #193, opened 2018-10-02
Fix missing swagger params for delete

sethisernhagen:
- Fix missing swagger query string params for delete operation

Why?

Delete operations with parameters were missing parameters in
its swagger definition.


microcosm_flask/conventions/crud.py
```diff
@@ -217,6 +217,7 @@ def configure_delete(self, ns, definition):
         request_schema = definition.request_schema or Schema()
 
         @self.add_route(ns.instance_path, Operation.Delete, ns)
+        @qs(definition.request_schema)
```
netama: I think you want to add `or Schema()` like what we do in https://github.com/globality-corp/microcosm-flask/commit/eb8f83d8547e125f91622daf0d609f4bdd676e62#diff-d021575c22812976dcde670b51504493R157, otherwise this might break when a schema is not specified (which is the common case). maybe add unittests for both cases to make sure?

sethisernhagen: I'm sure I'm not understanding. `or Schema()` is on line 217 ^.

netama: oops, missed that. so just use `request_schema` instead of `definition.request_schema`?

PR #198, opened 2018-10-08
GLOB-25917: Make sure that a statsd client is configured

moshez:
Why?

If `metrics_counting`/`metrics_timing` are going to be called,
we need to make sure that they have a statsd client on the graph
available as `graph.metrics`. Since the latest release of `microcosm_metrics`
only supports the DataDog client API, configure that right away
instead of hoping it will get taken care of at the application level.


moshez:
Closing this.

PR #200, opened 2018-10-11
Separate route path construction into its own component

jessemyers:
Path construction is difficult to modify today, in part because the `make_path` function
is not exposed and is not customizable.

This PR takes the first step of putting path construction into a graph component, adding
more exhaustive tests, and improving (read: regressing) a few behaviors. Customization
is possible via a `prefix` override parameter, but the builder cannot be swapped out yet.

There are two backwards imcompatible changes here:

 - `Namespace.prefix` is no longer used for the namespace's own path calculation; it is instead
   used by the `RoutePathBuilder` component. By inspection, there are no obvious usages of this
   parameter in our code any more (we use `version` instead), so this should be safe.

 - Trailing slashes are removed from paths. This means that the current discovery convention
   that handles `/api/` will handle `/api` instead. No code should depend on this endpoint and
   the new usage should be better for humans.

### DIFF
--- a/microcosm_flask/tests/conventions/test_discovery.py
+++ b/microcosm_flask/tests/conventions/test_discovery.py
@@ -27,7 +27,7 @@ def test_discovery():
 
     client = graph.flask.test_client()
 
-    response = client.get("/api/")
+    response = client.get("/api")
     assert_that(response.status_code, is_(equal_to(200)))
     data = loads(response.get_data().decode("utf-8"))
     assert_that(data, is_(equal_to({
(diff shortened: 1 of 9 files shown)


microcosm_flask/tests/conventions/test_discovery.py
```diff
@@ -27,7 +27,7 @@ def search_foo():
 
     client = graph.flask.test_client()
 
-    response = client.get("/api/")
+    response = client.get("/api")
```
jessemyers: Notice me.

microcosm_flask/tests/test_paths.py
```diff
@@ -0,0 +1,62 @@
...
+            is_(equal_to("/api")),
+        )
+
+    def test_override_prefix(self):
+        assert_that(
+            self.build_route_path("/ping", "/foo"),
+            is_(equal_to("/foo/ping")),
+        )
+
+    def test_override_slash_prefix(self):
```
jessemyers: Notice me @piercefreeman

PR #204, opened 2018-12-10
Fix swagger def generation for no version/no prefix case

afallou:
**Why?**
Though it's not something we've done before, one could want to have non-namespaced paths, i.e. all paths under the root without a prefix or version number.
One example is when paths are imposed by a third-party tool that will contact our API on a fixed path (e.g. building a container for AWS SageMaker).

Since a path suffix has a leading `/`, when the API prefix and version number are both empty we'd expect the base path to equal `""`. However, OpenAPI requires the base path to minimally be `"/"`, which requires a special case to generate a proper OpenAPI definition.

**What?**
- Fix paths in the OpenAPI definition to have a leading slash when the API prefix and version are empty strings.
- Add test for that particular case.

### DIFF
--- a/microcosm_flask/swagger/definitions.py
+++ b/microcosm_flask/swagger/definitions.py
@@ -80,8 +80,10 @@ def add_paths(paths, base_path, operations):
         if not path.startswith(base_path):
             continue
         method = operation.value.method.lower()
+        # if there's no version number or prefix, then the root slash is also part of the path suffix
+        suffix_start = 0 if len(base_path) == 1 else len(base_path)
         paths.setdefault(
-            path[len(base_path):],
+            path[suffix_start:],
             swagger.PathItem(),
         )[method] = build_operation(operation, ns, rule, func)
 
--- a/microcosm_flask/swagger/definitions.py
+++ b/microcosm_flask/swagger/definitions.py
@@ -80,7 +80,9 @@ def add_paths(paths, base_path, operations):
         if not path.startswith(base_path):
             continue
         method = operation.value.method.lower()
-        # if there's no version number or prefix, then the root slash is also part of the path suffix
+        # If there is no version number or prefix, we'd expect the base path to be ""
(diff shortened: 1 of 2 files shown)


microcosm_flask/swagger/definitions.py
```diff
@@ -80,8 +80,10 @@ def add_paths(paths, base_path, operations):
         if not path.startswith(base_path):
             continue
         method = operation.value.method.lower()
+        # if there's no version number or prefix, then the root slash is also part of the path suffix
+        suffix_start = 0 if len(base_path) == 1 else len(base_path)
         paths.setdefault(
-            path[len(base_path):],
+            path[suffix_start:],
```
afallou: As @pokey mentioned in IRL discussion, this special-casing seems weird. I tend to agree, I'm going to dig deeper.

afallou: After digging a little - logically we'd want the base path to be `""` if both prefix and version are empty. However, an the base path in the OpenAPI spec must have a leading slash, so an empty base path is not possible; we'll have to keep the branching logic here.

jessemyers: Seems like you should at least make your condition be `base_path == "/"` ?

jessemyers: Or better still, express "I don't want a base path" as `None` and handle this a little more clearly, with something like:

```
if base_path is None:
    base_path = "/"
```

I'd need to look at the usage from a wider lens to get this right logically, but I think the gist should be:
 - The `Namespace` should support a usage that expresses what you want. I believe you want "no prefix" and that should mean `None` (and not `""`, even though that's "nicer" for the `path.startswith(base_path)` check on 80).
 - The business logic here should be abstracted out to a helper function or two with unit tests that operate at a lower level than "generate swagger"
 - This business logic should have a clearly delineated branch on the "no prefix" case and then document where it injects/enforces a prefix of "/"

PR #209, opened 2019-03-05
Refactor swagger parameter/field generation

jessemyers:
Key changes:
 -  Adds a dedicated package and class hierarchy for building swagger parameters
    from marshmallow fields. Breaks functionality into five different implementations
    with a unified building `@classmethod` that discovers registered builders from
    entry points.

 -  Refactors tests into parallel structure to builders.

 -  Moves `swagger_field`, `build_schema`, and `build_parameter` functions to more
    obvious locations (e.g. `decorators.py` and `api.py`)


microcosm_flask/swagger/api.py
```diff
@@ -0,0 +1,38 @@
...
+from microcosm_flask.swagger.parameters.base import ParameterBuilder
+from microcosm_flask.swagger.schema import iter_fields
+
+
+def build_schema(schema: Schema) -> Mapping:
+    """
+    Build JSON schema from a marshmallow schema.
+
+    """
+    fields = list(iter_fields(schema))
```
jessemyers: It will eventually make sense to have a `SchemaBuilder` to mirror the `ParameterBuilder` structure. We don't yet have the same need for flexibility, but it'd make the code easier to follow.

microcosm_flask/swagger/decorators.py
```diff
@@ -0,0 +1,16 @@
+from microcosm_flask.swagger.parameters.decorated import SWAGGER_FORMAT, SWAGGER_TYPE
+
+
+def swagger_field(field, swagger_type="string", swagger_format=None):
```
jessemyers: I found one usage of this "decorator" in our code. It's also not a proper decorator, which is something I'd like to fix. This comes at some expense to backwards compatibility, but I think it's justified.

microcosm_flask/swagger/parameters/base.py
```diff
@@ -0,0 +1,142 @@
...
+from marshmallow.fields import Field, List, Nested
+
+from microcosm_flask.naming import name_for
+from microcosm_flask.swagger.naming import type_name
+
+
+ENTRY_POINT = "microcosm_flask.swagger.parameters"
+
+
+class ParameterBuilder(ABC):
```
jessemyers: Most of the work is in this base class.

microcosm_flask/swagger/parameters/base.py
```diff
@@ -0,0 +1,142 @@
...
+        }
+
+    @classmethod
+    def build(cls, field: Field) -> Mapping:
+        """
+        Build a swagger parameter from a marshmallow field.
+
+        """
+        builders = [
+            entry_point.load()()
```
jessemyers: We can customize how we generate fields by registering a new entry point (e.g. in some other code base)

microcosm_flask/swagger/parameters/base.py
```diff
@@ -0,0 +1,142 @@
...
+
+        return builder.build_parameter(field)
+
+    @classmethod
+    def default_builder_type(cls):
+        """
+        Define the default builder type.
+
+        """
+        # avoid circular dependencies
```
jessemyers: I don't love this, but it turns out that some kinds of builders (e.g. lists) need the ability to reference the top-level `build` function directly. This means we either need to pass the builder (or a `graph`) around of ensure that every builder can call itself to `build` (via the above `@classmethod`). I went with the latter due to none of this code using the `graph` yet. This comes with the problem that we can't import the default implementation in the usual way because the base class has to finish loading first.

microcosm_flask/swagger/parameters/decorated.py
```diff
@@ -0,0 +1,29 @@
...
+from marshmallow.fields import Field
+
+from microcosm_flask.swagger.parameters.base import ParameterBuilder
+
+
+SWAGGER_TYPE = "__swagger_type__"
+SWAGGER_FORMAT = "__swagger_format__"
+
+
+class DecoratedParameterBuilder(ParameterBuilder):
```
jessemyers: One of the motivations for this PR is the ability to extend the decorated builder to do more things. The surface area for such changes is now small enough that this is sane.

microcosm_flask/swagger/decorators.py
GabeLambda: If it's in [argus](https://github.com/globality-corp/argus/blob/70ae4f97a5881fd0ede70e826f5608e2b28d77ab/argus/resources/troubleshooting_resources.py#L12), then we could probably drop it.

microcosm_flask/swagger/parameters/default.py
```diff
@@ -0,0 +1,61 @@
...
+class DefaultParameterBuilder(ParameterBuilder):
+    """
+    Builder parameters using defaults field mappings.
+
+    """
+    def supports_field(self, field: Field) -> bool:
+        try:
+            FIELD_MAPPINGS[type(field)]
+        except KeyError:
+            self.logger.exception(  # type: ignore
```
GabeLambda: You can also declare a `logger` class attribute, and not assign it anything.

microcosm_flask/swagger/decorators.py
jessemyers: I removed the `argus` usage, which leaves me free to make this work the way it should. Nice.

microcosm_flask/swagger/parameters/base.py
```diff
@@ -0,0 +1,142 @@
...
+    def parse_ref(self, field: Field) -> Optional[str]:
+        """
+        Parse the reference type for nested fields, if any.
+
+        """
+        if not isinstance(field, Nested):
+            return None
+
+        ref_name = type_name(name_for(field.schema))
+        return f"#/definitions/{ref_name}"
```
pokey: I'd be tempted to put this and the above into separate builders

microcosm_flask/swagger/parameters/base.py
```diff
@@ -0,0 +1,142 @@
...
+from pkg_resources import iter_entry_points
+from typing import Any, Iterable, Mapping, Optional, Sequence, Tuple
+
+from marshmallow.fields import Field, List, Nested
+
+from microcosm_flask.naming import name_for
+from microcosm_flask.swagger.naming import type_name
+
+
+ENTRY_POINT = "microcosm_flask.swagger.parameters"
```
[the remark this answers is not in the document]
jessemyers: I was trying to mirror that actual package structure.

microcosm_flask/swagger/parameters/base.py
```diff
@@ -0,0 +1,142 @@
...
+    """
+    def __init__(self):
+        self.parsers = {
+            "default": self.parse_default,
+            "description": self.parse_description,
+            "enum": self.parse_enum_values,
+            "format": self.parse_format,
+            "$ref": self.parse_ref,
+            "type": self.parse_type,
+            "items": self.parse_items,
```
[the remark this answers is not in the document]
jessemyers: Yeah, one of those. Given that `minimum` and `maximum` are potentially useful in general, adding hooks here would make sense (and then extending the *decorated* builder accordingly)

microcosm_flask/swagger/parameters/base.py
jessemyers: It would, but it pretty much means introducing `graph` because there are recursive calls to this function so we need a back-reference.

This is probably the right thing to do.

microcosm_flask/swagger/parameters/base.py
jessemyers: Good idea.

microcosm_flask/swagger/parameters/default.py
```diff
@@ -0,0 +1,64 @@
...
+from marshmallow.fields import Field
+from microcosm_logging.decorators import logger
+
+from microcosm_flask.fields import EnumField, LanguageField, QueryStringList, URIField
+from microcosm_flask.swagger.parameters.base import ParameterBuilder
+
+
+# see: https://github.com/marshmallow-code/apispec/blob/dev/apispec/ext/marshmallow/swagger.py
+FIELD_MAPPINGS = {
+    EnumField: (None, None),
```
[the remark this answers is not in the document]
jessemyers: Agreed. This was copy-pasted from before.

microcosm_flask/swagger/parameters/base.py
pokey: I don't see why we can't just split this into two classes, one which is a base class for all builders, and the other which is basically a registry for builders that selects which one to use

microcosm_flask/swagger/parameters/base.py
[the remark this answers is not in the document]
pokey: Oh I see.  Could go either way then

PR #211, opened 2019-04-15
Add a CRUDStoreAdapter specifically for encryptable models

GabeLambda:
Currently only adds an override to support encrypted model re-encryption
(e.g. for re-encrypting data using a separate master key).

I haven't yet added any tests, I wanted feedback on basic structure before I started checking anything here.

### DIFF
--- a/microcosm_flask/encryption/conventions/crud_adapter.py
+++ b/microcosm_flask/encryption/conventions/crud_adapter.py
@@ -7,7 +7,7 @@ class EncryptableCRUDStoreAdapter(CRUDStoreAdapter):
 
     """
 
-    def update(self, **kwargs):
+    def update_and_reencrypt(self, **kwargs):
         """
         Support re-encryption by enforcing that every update triggers a
         new encryption call, even if the the original call does not update
--- a/microcosm_flask/encryption/conventions/crud_adapter.py
+++ b/microcosm_flask/encryption/conventions/crud_adapter.py
@@ -1,0 +1,41 @@
+from microcosm_flask.conventions.crud_adapter import CRUDStoreAdapter
+
+
+class EncryptableCRUDStoreAdapter(CRUDStoreAdapter):
+    """
+    Adapt the CRUD conventions callbacks to the `EncryptableStore` interface.
+
+    """
+
+    def update(self, **kwargs):
+        """
+        Support re-encryption by enforcing that every update triggers a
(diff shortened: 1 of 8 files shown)


microcosm_flask/encryption/conventions/crud_adapter.py
```diff
@@ -0,0 +1,41 @@
...
+
+        """
+        encrypted_field_name = self.store.model_class.__plaintext__
+
+        id_ = kwargs[self.identifier_key]
+        current_model = self.store.retrieve(id_)
+        current_value = current_model.plaintext
+
+        null_update = (
+            # Check if the update is for the encrypted field, and if it's explicitly set to null
```
pelegdvir: That's going to be problematic  - with default values:
```
update(self, foo=None, **kwargs):
    return super().update(foo=foo, **kwargs)
```

microcosm_flask/encryption/conventions/crud_adapter.py
```diff
@@ -0,0 +1,41 @@
+from microcosm_flask.conventions.crud_adapter import CRUDStoreAdapter
+
+
+class EncryptableCRUDStoreAdapter(CRUDStoreAdapter):
+    """
+    Adapt the CRUD conventions callbacks to the `EncryptableStore` interface.
+
+    """
+
+    def update(self, **kwargs):
```
pelegdvir: Maybe it's preferable to call this: update_and_reencrypt.
In some cases, we want to use the update function again and again without reencrypting the data.

GabeLambda: Okay.

microcosm_flask/encryption/conventions/crud_adapter.py
GabeLambda: I'd argue that updates shouldn't ever pass in default values. You should only be specifying exactly what you want to update here.

microcosm_flask/tests/encryption/conventions/fixtures/encryptable_models.py
```diff
@@ -0,0 +1,24 @@
...
+from uuid import UUID
+
+
+@dataclass
+class Encryptable:
+    id: UUID
+    value: Optional[str] = None
+    other_value: Optional[str] = None
+
+    # # encryption support
```
pelegdvir: ?

PR #214, opened 2019-05-13
Control request header logging.

cdm-ium:
We have two cases in microcosm-flask for inspecting request headers:
- In the audit log we want to log the appropriate Globality-specific
  headers.
- In the resulting actions of the web-route we want to pass along the
  appropriate headers in `graph.opaque`.

It turns out that these aren't always the same. In particular we define
two different sets of headers:

- `X-Request`: These are headers already used within microcosm for
tracking events across an entire user flow which may include *many* http
requests.
- `X-Client`: These are hreaders for tracking information across one
single HTTP request, usually only involving a single client and a single
server.

Before this PR we used the `graph` abstraction for getting `request_context`, but on looking at our code we only use `graph.request_context` from within this repo, so I decided to simplify the code a bit not to use the graph.

### DIFF
--- a/microcosm_flask/context.py
+++ b/microcosm_flask/context.py
@@ -1,5 +1,6 @@
 from flask import request
 
+
 # This is the total set of Globality-proprietary headers
 ALL_HEADER_LIST = (
     "X-Request",  # These are values *invariant* to a user request, e.g. user-id, request-id, etc.
--- a/microcosm_flask/context.py
+++ b/microcosm_flask/context.py
@@ -1,13 +1,14 @@
 from flask import request
 
 
-# This is the total set of Globality-proprietary headers
+# Special headers reserved for use in `microcosm-` based applications:
 ALL_HEADER_LIST = (
-    "X-Request",  # These are values *invariant* to a user request, e.g. user-id, request-id, etc.
+    "X-Request",  # These are values *invariant* to a user request from UI
+                  # perspective, e.g. user-id, request-id, etc.
     "X-Client",  # These are values *specific* to a single HTTP Client call to a web service
 )
 
-# This is the list of internal Globality Headers we use which are safe for logging.
+# This is the list of internal Headers which should always be logged.
 LOGGABLE_HEADER_WHITE_LIST = ("X-Request")
 
 
--- a/microcosm_flask/context.py
+++ b/microcosm_flask/context.py
@@ -10,10 +10,12 @@ ALL_HEADER_LIST = (
 LOGGABLE_HEADER_WHITE_LIST = ("X-Request")
 
 
-def get_request_context(header_white_list=LOGGABLE_HEADER_WHITE_LIST):
-    return {
-        header: value
-        for prefix in header_white_list
-        for header, value in request.headers.items()
-        if header.startswith(prefix)
-    }
+def make_get_request_context(header_white_list=LOGGABLE_HEADER_WHITE_LIST):
+    def get_request_context():
+        return {
+            header: value
+            for prefix in header_white_list
+            for header, value in request.headers.items()
+            if header.startswith(prefix)
+        }
+    return get_request_context
--- a/microcosm_flask/context.py
+++ b/microcosm_flask/context.py
@@ -1,32 +1,19 @@
 from flask import request
-from microcosm.api import defaults
 
-
-X_REQUEST = "X-Request"
-
-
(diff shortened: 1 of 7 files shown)


microcosm_flask/tests/test_context.py
```diff
@@ -5,16 +5,32 @@
...
 
     with graph.flask.test_request_context(headers={
             "X-Request-Id": "foo",
     }):
-        with graph.opaque.initialize(graph.request_context):
+        with graph.opaque.initialize(make_get_request_context()):
             assert_that(graph.opaque["X-Request-Id"], is_(equal_to("foo")))
+
+
+def test_make_get_request_context_respects_passed_headers():
```
cdm-ium: This is important as it explains why this function is useful.

microcosm_flask/context.py
```diff
@@ -1,32 +1,22 @@
...
-def configure_request_context(graph):
-    """
-    Configure the flask context function which controls what data you want to associate
-    with your flask request context, e.g. headers, request body/response.
-
-    Usage:
-        graph.request_context()
-
-    """
-    include_header_prefix = graph.config.request_context.include_header_prefix
```
GabeLambda: Looks like we lose this config?

microcosm_flask/context.py
```diff
@@ -1,32 +1,22 @@
 from flask import request
-from microcosm.api import defaults
 
 
-X_REQUEST = "X-Request"
+# This is the total set of Globality-proprietary headers
```
GabeLambda: Are headers actually proprietary?

microcosm_flask/tests/test_context.py
```diff
@@ -5,16 +5,32 @@
...
+    graph.use(
+        "opaque",
+    )
+
+    with graph.flask.test_request_context(headers={
+            "X-Foo-Id": "foo",
+            "X-Bar-Id": "bar",
+    }):
+        with graph.opaque.initialize(make_get_request_context(["X-Bar"])):
+            assert_that(graph.opaque["X-Bar-Id"], is_(equal_to("bar")))
```
GabeLambda: I'm probably confused, why is one prefix working here? Neither of these are in the whitelist or the all_header list?

microcosm_flask/context.py
cdm-ium: Yes because we have never used it in over > 4 years and I think changing the header prefix requires thoughts about the design of the rest of the application. We could add it back, when we (or someone who depends on us) actually uses it.

microcosm_flask/tests/test_context.py
cdm-ium: `make_get_request_context` takes a list of `header_prefixes` to pull from the `request.headers`, so on the line above I say give me all the `X-Bar` headers which means those and only those headers will come back.

microcosm_flask/context.py
cdm-ium: `proprietary` is the wrong word, but these are headers `Globality` and consequently `microcosm` have opinions about. The `X-` part is a HTTP standard, but `X-Request` and `X-Client` are things we made up.

cdm-ium: Also adding: https://globality.atlassian.net/browse/GLOB-32485?oldIssueView=true for a more hollistic look at context in microcosm.

microcosm_flask/context.py
cdm-ium: Basically what I found while doing this is the config we have is insufficient. I think if we wanted to make this available through the graph we'd want something smarter which supports getting the right type of context depending on various factors (e.g. for a given user flow, a given http request, a given handler execution).

microcosm_flask/context.py
```diff
@@ -1,32 +1,23 @@
 from flask import request
-from microcosm.api import defaults
 
 
-X_REQUEST = "X-Request"
+# Special headers reserved for use in `microcosm-` based applications:
+ALL_HEADER_LIST = (
+    "X-Request",  # These are values *invariant* to a user request from UI
+                  # perspective, e.g. user-id, request-id, etc.
+    "X-Client",  # These are values *specific* to a single HTTP Client call to a web service
```
GabeLambda: Uber-nit: vertical `#`.

microcosm_flask/tests/test_context.py
GabeLambda: Ohh, I was missing L34. Got it.

PR #215, opened 2019-06-26
Allow audit logging at debug.

cdm-ium:
When we set out to add audit logging to microcosm we had a more naive
definition of audit logs. I think what we call "audit logs" are just
"request logs" and should be revisted later. For now, we need the
ability to turn some of these automatic logs off by lowering the logging
level to clean up our logs.

### DIFF
--- a/microcosm_flask/audit.py
+++ b/microcosm_flask/audit.py
@@ -13,7 +13,8 @@ from uuid import UUID
 
 from flask import current_app, g, request
 from inflection import underscore
-from microcosm.api import defaults
+from microcosm.api import defaults, typed
+from microcosm.config.types import boolean
 from microcosm_logging.timing import elapsed_time
 
 from microcosm_flask.errors import (
@@ -33,6 +34,7 @@ AuditOptions = namedtuple("AuditOptions", [
     "include_response_body",
     "include_path",
     "include_query_string",
+    "log_as_debug",
 ])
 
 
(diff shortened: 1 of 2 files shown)


microcosm_flask/audit.py
```diff
@@ -353,10 +358,11 @@ def parse_response(response):
 
 
 @defaults(
-    include_request_body=DEFAULT_INCLUDE_REQUEST_BODY,
-    include_response_body=DEFAULT_INCLUDE_RESPONSE_BODY,
-    include_path="true",
-    include_query_string="true",
+    include_request_body=typed(type=int, default_value=DEFAULT_INCLUDE_REQUEST_BODY),
```
cdm-ium: This has been bothering me for a while. We're manually doing various conversions farther down, but now we have code in microcosm to do this for us.

microcosm_flask/audit.py
```diff
@@ -183,7 +185,10 @@ def log(self, logger):
                 logger.warning(dct)
         else:
             # usually log at INFO; a raised exception can be an error or expected behavior (e.g. 404)
-            logger.info(self.to_dict())
+            if self.options.log_as_debug:
```
GabeLambda: Since the main path (default?) is probably going to be to log this at info, this could be inverted (you'll eat the branch misprediction every time otherwise).

GabeLambda: Might be a huge premature optimization.

cdm-ium: Not a big deal, besides optimization, I agree our brains expect the `if` to be the common case.

PR #216, opened 2019-07-10
Make request context capture jaeger headers

cdm-ium:
There only `jaeger` specific change is adding the `jaeger` header to the `microcosm-flask` whitelist of headers. Today the white list is a single header, so there is a generalization to support a list of prefixes.

### DIFF
--- a/microcosm_flask/context.py
+++ b/microcosm_flask/context.py
@@ -8,23 +8,23 @@ JAEGER_TRACE_ID = "Uber-Trace-Id"
 HEADER_PREFIXES = [X_REQUEST, JAEGER_TRACE_ID]
 
 
-def context_wrapper(include_header_prefix):
+def context_wrapper(include_header_prefixes):
     def retrieve_context():
         context = {
             header: value
             for header, value in request.headers.items()
             if any([
                 header.startswith(prefix)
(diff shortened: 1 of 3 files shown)


microcosm_flask/context.py
```diff
@@ -1,23 +1,29 @@
 from flask import request
 from microcosm.api import defaults
+from microcosm.tracing import TRACE_ID, SPAN_NAME
```
cdm-ium: other services know about support for `tracing` but they don't know anything `jaeger` specific.

cdm-ium:
Build will fail until `microcosm` piece is released: https://github.com/globality-corp/microcosm/pull/51/files

microcosm_flask/context.py
```diff
@@ -1,23 +1,29 @@
...
             header: value
             for header, value in request.headers.items()
-            if header.startswith(include_header_prefix)
+            if any([
+                header.lower().startswith(prefix.lower())
+                for prefix in include_header_prefixes
+            ])
         }
-
+        context[SPAN_NAME] = str(request)
```
GabeLambda: I'd make a similar comment here about making this optional, not sure if its worth the time.

PR #217, opened 2019-07-11
Pyinstrument profiler

salmanmashayekh:


setup.py
```diff
@@ -30,6 +30,7 @@
         "python-dateutil>=2.7.3",
         "PyYAML>=3.13",
         "rfc3986>=1.2.0",
```
adamhadani: LGTM one caveat - I would add pyinstrimenr under extras_required as “profiling”. You may need to them try catch importing it in the code.

salmanmashayekh: @adamhadani I am not too familiar with `extras_require`, but I gave it a shot.

microcosm_flask/profiling.py
```diff
@@ -1,7 +1,14 @@
...
+from os.path import exists, expanduser, join
 
-from werkzeug.contrib.profiler import ProfilerMiddleware
+from flask import g
+
+
+try:
+    from pyinstrument import Profiler
+except ImportError:
+    raise Exception("Profiling requires 'pyinstrument'.")
```
adamhadani: the purpose with catching import error is actually to swallow the exception so that if you dont have the extras_required, just warn the user if at all. so in this case maybe as simple as:

```python
...
except ImportError:
    pass
```

and then can add a check below inside the function along lines of:

```python
...
  if not getattr(g, 'profiler', None):
    raise Exception("profiler is not installed. make sure pyinstrument is installed")
```

salmanmashayekh: I think it is good to go now @adamhadani !

PR #225, opened 2019-08-12
Add memory profiling support.

cdm-ium:
Why? We have excellent CPU profiling support, but no memory support.
Recently we enocuntered a memory leak integrating another third party
library and found ourselves blind to debugging the leak in our
environment.

This tooling is primitive but should be sufficient for identifying leaks
in the application. This approach is copied from:

https://superuser.blog/detect-memory-leak-python/

which uses a built-in python memory tracking library: `tracemalloc`.

I do have images of the logs this produces, but they use our internal/private services as an example so I don't want to post them here. If you work for Globality, you can see screenshots in #resilience-team.

### DIFF
--- a/microcosm_flask/memory.py
+++ b/microcosm_flask/memory.py
@@ -1,0 +1,41 @@
+from logging import Logger
+from tracemalloc import start, take_snapshot
+
+from microcosm.api import binding, defaults, typed
+from microcosm.config.types import boolean
+
+
+@binding("memory_profiler")
+@defaults(
+    enabled=typed(boolean, False),
+    report_size_lines=typed(int, 10),
+)
+class MemoryProfiler:
+    logger: Logger
+
+    def __init__(self, graph):
+        self.enabled = graph.config.memory_profiler.enabled
+        self.report_size_lines = graph.config.memory_profiler.report_size_lines
+        self.logger = graph.logger
+
+        if not self.enabled:
+            self.logger.info("Skipping initialization because memory profiling is not enabled!")
+            return
+
+        start()
+        self.origin_snapshot = take_snapshot()
+
+    def take_snapshot(self, func):
+        def wrapped(*args, **kwargs):
+            result = func(*args, **kwargs)
+            if not self.enabled:
+                self.logger.debug("Memory profiling is disabled. Please enable to get snapshots.")
+                return result
+
+            latest_snapshot = take_snapshot()
+            difference = latest_snapshot.compare_to(self.origin_snapshot, "lineno")
+            top_ten = difference[:self.report_size_lines]
+            worst_offenders = "\n".join([str(memory_item) for memory_item in top_ten])
+            self.logger.info(f"Memory snapshot: \n {worst_offenders}")
+            return result
+        return wrapped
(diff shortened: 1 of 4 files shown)


microcosm_flask/memory.py
```diff
@@ -0,0 +1,41 @@
...
+class MemoryProfiler:
+    logger: Logger
+
+    def __init__(self, graph):
+        self.enabled = graph.config.memory_profiler.enabled
+        self.report_size_lines = graph.config.memory_profiler.report_size_lines
+        self.logger = graph.logger
+
+        if not self.enabled:
+            self.logger.info("Skipping initialization because memory profiling is not enabled!")
```
cdm-ium: If disabled, there is a single, info level log which will tell you this.

microcosm_flask/memory.py
```diff
@@ -0,0 +1,41 @@
...
+            return
+
+        start()
+        self.origin_snapshot = take_snapshot()
+
+    def take_snapshot(self, func):
```
cdm-ium: I ebbed and flowed on using a decorator, but that's what fits the pattern we have for adding this behavior on a route in mc-flask. We can change/generalize if there's ever a use case.

microcosm_flask/memory.py
```diff
@@ -0,0 +1,41 @@
...
+        def wrapped(*args, **kwargs):
+            result = func(*args, **kwargs)
+            if not self.enabled:
+                self.logger.debug("Memory profiling is disabled. Please enable to get snapshots.")
+                return result
+
+            latest_snapshot = take_snapshot()
+            difference = latest_snapshot.compare_to(self.origin_snapshot, "lineno")
```
cdm-ium: Technically taking a diff may not be necessary to get interesting information, but this is orthogonal to the actual problem I'm trying to solve. Right now I just want to get something that works into the library. Future generations can improve.

microcosm_flask/memory.py
```diff
@@ -0,0 +1,41 @@
...
+            top_ten = difference[:self.report_size_lines]
+            worst_offenders = "\n".join([str(memory_item) for memory_item in top_ten])
+            self.logger.info(f"Memory snapshot: \n {worst_offenders}")
```
GabeLambda: I'm worried that this might take up a fair amount of logs. Can you limit this per X number of calls?

cdm-ium: I ended up doing it by time, so only once every X min.

microcosm_flask/tests/test_memory.py
```diff
@@ -0,0 +1,56 @@
...
+        self.graph = create_object_graph("example", testing=True, debug=True, loader=self.loader)
+        self.graph.use(
+            "flask",
+            "memory_profiler",
+        )
+
+        self.now = datetime.now()
+        self.last_sampling_time = self.now - timedelta(minutes=5)
+        self.graph.memory_profiler.last_sampling_time = self.last_sampling_time
+        print(self.graph.memory_profiler.last_sampling_time_delta)
```
GabeLambda: print?

pokey:
nice!  so were you able to find the memory leak?

PR #227, opened 2019-08-16
update test for new microcosm classifier behavior

clatko-globe:
test will only pass against version of microcosm-metrics sitting in pr - https://github.com/globality-corp/microcosm-metrics/pull/15
ref https://globality.atlassian.net/browse/DEVOPS-347

### DIFF
--- a/microcosm_flask/tests/test_metrics.py
+++ b/microcosm_flask/tests/test_metrics.py
@@ -57,13 +57,15 @@ class TestRouteMetrics:
             tags=[
                 "endpoint:foo.search.v1",
                 "backend_type:microcosm_flask",
+                "classifier:200",
             ],
         )
         self.graph.metrics.increment.assert_called_with(
-            "route.200.count",
+            "route.call.count",
             tags=[
                 "endpoint:foo.search.v1",
                 "backend_type:microcosm_flask",
+                "classifier:200",
             ],
         )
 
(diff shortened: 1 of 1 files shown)


microcosm_flask/tests/test_metrics.py
```diff
@@ -57,13 +57,15 @@ def foo():
...
+                "classifier:200",
             ],
         )
         self.graph.metrics.increment.assert_called_with(
-            "route.200.count",
+            "route.call.count",
             tags=[
                 "endpoint:foo.search.v1",
                 "backend_type:microcosm_flask",
+                "classifier:200",
```
cdm-ium: Reading this, what is the meaning of `classifier`? When is it not `status_code`? What are other uses of `classifier` that justify the more generic name? Seems like we should just have more, more specific tags.

amenut: If the label here is for `status_code`, call it `status_code`.
Better synecdoche for when you (or anyone else) comes along later and is writing PromQL queries around status codes.

clatko-globe: the reason for the naming is derived from microcosm metrics - https://github.com/globality-corp/microcosm-metrics#decorators - you pass a classifier that implements `label_result` and `label_error` and that is used to comprise part of the key. i should not make this more specific as we may want this to hold other types of information. i do not want to change the method signature for compatibility. I'll add comments on why the name is what it is.

GabeLambda: I still think that we can do better about labeling here (`classifier` is needlessly obtuse). But we can tackle that another time.

PR #231, opened 2019-08-19
GLOB-35427 Upgrade to marshmallow 3.0

GabeLambda:
Updates usages of marshmallow to account for several changes in marshmallow 3.0

Note that this effectively moves this repo to be python 3.5+ compatible-only, given that those are the versions that marshmallow itself supports.

Note: I haven't run this in another live service yet; I wouldn't be surprised if there are other subtle issues that may have cropped up here. I hadn't yet pored through every line in the [upgrade doc](https://marshmallow.readthedocs.io/en/stable/upgrading.html). I'll do some testing and read through that first before merging.

### DIFF
--- a/microcosm_flask/conventions/crud.py
+++ b/microcosm_flask/conventions/crud.py
@@ -288,8 +288,7 @@ class CRUDConvention(Convention):
         @wraps(definition.func)
         def update(**path_data):
             headers = dict()
-            # NB: using partial here means that marshmallow will not validate required fields
-            request_data = load_request_data(definition.request_schema, partial=True)
+            request_data = load_request_data(definition.request_schema)
             response_data = require_response_data(definition.func(**merge_data(path_data, request_data)))
             definition.header_func(headers, response_data)
             response_format = self.negotiate_response_content(definition.response_formats)
@@ -320,9 +319,7 @@ class CRUDConvention(Convention):
         @wraps(definition.func)
         def create_collection(**path_data):
             request_data = load_request_data(definition.request_schema)
-            # NB: if we don't filter the request body through an explicit page schema,
-            # we will leak other request arguments into the pagination query strings
-            page = self.page_cls.from_query_string(self.page_schema(), request_data)
+            page = self.page_cls.from_query_string(self.page_schema(), {})
 
             result = definition.func(**merge_data(
                 path_data,
--- a/microcosm_flask/conventions/encoding.py
+++ b/microcosm_flask/conventions/encoding.py
@@ -97,7 +97,13 @@ def load_query_string_data(request_schema, query_string_data=None):
     if query_string_data is None:
         query_string_data = request.args
 
-    return request_schema.load(query_string_data)
+    try:
+        return request_schema.load(query_string_data)
+    except ValidationError as error:
+        raise with_context(
+            UnprocessableEntity("Validation error"),
+            error.messages,
+        )
 
 
 def remove_null_values(data):
--- a/microcosm_flask/conventions/encoding.py
+++ b/microcosm_flask/conventions/encoding.py
@@ -4,6 +4,7 @@ Support for encoding and decoding request/response content.
 """
 from flask import request
 from inflection import camelize
+from marshmallow.exceptions import ValidationError
 from werkzeug.exceptions import NotFound, UnprocessableEntity
 
 from microcosm_flask.enums import ResponseFormats
@@ -55,7 +56,7 @@ def encode_headers(resource):
     return {}
 
 
-def load_request_data(request_schema, partial=False):
+def load_request_data(request_schema):
     """
     Load request data as JSON using the given schema.
 
@@ -68,20 +69,22 @@ def load_request_data(request_schema, partial=False):
     try:
         json_data = request.get_json(force=True) or {}
     except Exception:
-        # if `simplpejson` is installed, simplejson.scanner.JSONDecodeError will be raised
+        # if `simplejson` is installed, simplejson.scanner.JSONDecodeError will be raised
         # on malformed JSON, where as built-in `json` returns None
         json_data = {}
-    request_data = request_schema.load(json_data, partial=partial)
-    if request_data.errors:
-        # pass the validation errors back in the context
+    try:
+        return request_schema.load(json_data)
+    except ValidationError as error:
         raise with_context(
-            UnprocessableEntity("Validation error"), [{
-                "message": "Could not validate field: {}".format(field),
-                "field": field,
-                "reasons": reasons
-            } for field, reasons in request_data.errors.items()],
+            UnprocessableEntity("Validation error"),
+            [
+                {
+                    "message": "Could not validate field: {}".format(field),
+                    "field": field,
+                    "reasons": reasons,
+                } for field, reasons in error.messages.items()
+            ],
         )
-    return request_data.data
 
 
 def load_query_string_data(request_schema, query_string_data=None):
@@ -94,11 +97,7 @@ def load_query_string_data(request_schema, query_string_data=None):
     if query_string_data is None:
         query_string_data = request.args
 
-    request_data = request_schema.load(query_string_data)
-    if request_data.errors:
-        # pass the validation errors back in the context
-        raise with_context(UnprocessableEntity("Validation error"), dict(errors=request_data.errors))
-    return request_data.data
+    return request_schema.load(query_string_data)
 
 
 def remove_null_values(data):
@@ -128,7 +127,7 @@ def dump_response_data(response_schema,
 
     """
     if response_schema:
-        response_data = response_schema.dump(response_data).data
+        response_data = response_schema.dump(response_data)
 
     return make_response(response_data, response_schema, response_format, status_code, headers)
 
--- a/microcosm_flask/fields/enum_field.py
+++ b/microcosm_flask/fields/enum_field.py
@@ -27,7 +27,7 @@ class EnumField(Field):
         self.by_value = by_value
         super(EnumField, self).__init__(*args, **kwargs)
 
-    def _serialize(self, value, attr, obj):
+    def _serialize(self, value, attr, obj, **kwargs):
         if value is None:
             return value
         elif isinstance(value, str) and not isinstance(value, Enum):
@@ -37,7 +37,7 @@ class EnumField(Field):
         else:
             return value.name
 
-    def _deserialize(self, value, attr, data):
+    def _deserialize(self, value, attr, data, **kwargs):
         if value is None:
             return value
         elif self.by_value:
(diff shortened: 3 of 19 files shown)


microcosm_flask/conventions/crud.py
```diff
@@ -288,8 +288,7 @@ def configure_update(self, ns, definition):
         @wraps(definition.func)
         def update(**path_data):
             headers = dict()
-            # NB: using partial here means that marshmallow will not validate required fields
-            request_data = load_request_data(definition.request_schema, partial=True)
+            request_data = load_request_data(definition.request_schema)
```
GabeLambda: I'm not sure why we wouldn't want to validate required fields on update; this is for the request itself, and if we specified required fields there, we should be respecting that.

microcosm_flask/conventions/crud.py
```diff
@@ -320,9 +319,7 @@ def configure_createcollection(self, ns, definition):
         @wraps(definition.func)
         def create_collection(**path_data):
             request_data = load_request_data(definition.request_schema)
-            # NB: if we don't filter the request body through an explicit page schema,
-            # we will leak other request arguments into the pagination query strings
-            page = self.page_cls.from_query_string(self.page_schema(), request_data)
+            page = self.page_cls.from_query_string(self.page_schema(), {})
```
GabeLambda: This was a tad confusing to me. I removed this because it started throwing up validation errors because it was trying to load `request_data` fields as query fields, which it was failing at.

I'm probably misunderstanding why we'd want "page" in the first place.

microcosm_flask/conventions/encoding.py
```diff
@@ -68,20 +69,22 @@ def load_request_data(request_schema, partial=False):
...
-        # if `simplpejson` is installed, simplejson.scanner.JSONDecodeError will be raised
+        # if `simplejson` is installed, simplejson.scanner.JSONDecodeError will be raised
         # on malformed JSON, where as built-in `json` returns None
         json_data = {}
-    request_data = request_schema.load(json_data, partial=partial)
-    if request_data.errors:
-        # pass the validation errors back in the context
+    try:
+        return request_schema.load(json_data)
+    except ValidationError as error:
```
GabeLambda: I main different with 3.0 is that they treat data only as data, errors are explicit exceptions.

microcosm_flask/conventions/encoding.py
```diff
@@ -94,11 +97,7 @@ def load_query_string_data(request_schema, query_string_data=None):
     if query_string_data is None:
         query_string_data = request.args
 
-    request_data = request_schema.load(query_string_data)
-    if request_data.errors:
-        # pass the validation errors back in the context
-        raise with_context(UnprocessableEntity("Validation error"), dict(errors=request_data.errors))
-    return request_data.data
+    return request_schema.load(query_string_data)
```
GabeLambda: May want to wrap this here too; none of the tests complained, but that may be because there aren't any.

microcosm_flask/fields/enum_field.py
```diff
@@ -27,7 +27,7 @@ def __init__(self, enum, by_value=False, *args, **kwargs):
         self.by_value = by_value
         super(EnumField, self).__init__(*args, **kwargs)
 
-    def _serialize(self, value, attr, obj):
+    def _serialize(self, value, attr, obj, **kwargs):
```
GabeLambda: The base class now throws a bunch of extra kwargs in here; subclasses need to respect that signature.

microcosm_flask/swagger/parameters/list.py
```diff
@@ -18,7 +18,7 @@ def parse_items(self, field: Field) -> Mapping[str, Any]:
         Parse the child item type for list fields, if any.
 
         """
-        return self.build_parameter(field.container)
+        return self.build_parameter(field.inner)
```
GabeLambda: Tracing through the code, `container` was removed, and `inner` looks to be a nearly identical replacement.

microcosm_flask/swagger/schemas.py
```diff
@@ -62,7 +62,8 @@ def iter_fields(self, schema: Schema) -> Iterable[Tuple[str, Field]]:
         """
         for name in sorted(schema.fields.keys()):
             field = schema.fields[name]
-            yield field.dump_to or name, field
+            # XXX: dump_to was removed; it wasn't being used anywhere? double-check
```
GabeLambda: Pay attention to me.

microcosm_flask/tests/conventions/fixtures.py
```diff
@@ -123,11 +123,13 @@ def address_delete(address_id, address_clock):
 def address_search(offset, limit, list_param=None, enum_param=None):
     if list_param is None or enum_param is None:
         return [ADDRESS_1], 1
-    return Address(
-        ADDRESS_ID_1,
-        PERSON_ID_1,
-        ",".join(list_param) + str(len(list_param)) + enum_param.value
-    ), 1
+    return [
```
GabeLambda: This was a weird one to me: the tests assumed that we would auto-convert single objects to lists, but I'm not sure if any service that uses mc-flask used that behavior. This was also breaking, because now the validation was running on the object itself, given the lack of conversion.

microcosm_flask/tests/conventions/test_crud.py
```diff
@@ -69,7 +70,7 @@ def add_request_id(headers, response_data):
     Operation.Replace: (person_replace, NewPersonSchema(), PersonSchema()),
     Operation.Retrieve: (person_retrieve, PersonLookupSchema(), PersonSchema()),
     Operation.Search: (person_search, OffsetLimitPageSchema(), PersonSchema()),
-    Operation.Update: (person_update, NewPersonSchema(), PersonSchema()),
+    Operation.Update: (person_update, UpdatePersonSchema(), PersonSchema()),
```
GabeLambda: I don't know why we weren't using this schema.

microcosm_flask/tests/fields/test_language_field.py
```diff
@@ -26,33 +26,33 @@ class LanguageSchema(Schema):
...
         ))
         assert_that(
-            result.data["language"],
+            result["language"],
             is_(equal_to(case)),
         )
 
 
 def test_load():
-    schema = LanguageSchema(strict=True)
```
GabeLambda: strict was removed, it's now the default.

microcosm_flask/conventions/crud.py
cdm-ium: I have no idea. It seems intentional, but don't remember why.

microcosm_flask/tests/test_paging.py
```diff
@@ -42,12 +44,19 @@ def test_offset_limit_page_to_from_dict():
...
-        # schema filters out extra arguments
-        assert_that(page.to_dict(), is_not(has_entry("foo", "bar")))
+
+
+def test_offset_limit_page_rejects_unknown_queries():
+    graph = create_object_graph(name="example", testing=True)
+    with graph.flask.test_request_context(query_string="offset=1&foo=bar"):
+        assert_that(
+            calling(OffsetLimitPage.from_query_string).with_args(OffsetLimitPageSchema()),
+            raises(ValidationError),
```
GabeLambda: Note: This is a *breaking* change; we now validate all incoming query arguments explicitly.

microcosm_flask/conventions/crud.py
cdm-ium: so it looks like this is trying to support query params in batch create operations. In theory this seems like it could be useful if the client somehow posts elements from a batch in pages/or if we use query params to control things like inferring fields of nested child schemas from the parent. I'm not sure the former is a real thing. I know we do the latter, but I'm not sure if we control the behavior with query params.

If you are going to run the `page_schema` against and empty dict, you may as well not run this line at all as it will never get data.

microcosm_flask/conventions/encoding.py
cdm-ium: It seems like you should.

microcosm_flask/swagger/schemas.py
cdm-ium: I think you're safe: https://github.com/search?q=org%3Aglobality-corp+dump_to&type=Code I only see it used in this repo, on this line.

microcosm_flask/tests/conventions/fixtures.py
cdm-ium: I don't understand what the `list_param` or `enum_param` args or for. Were these things we thought we wanted to use in microcosm that we never did?

microcosm_flask/tests/test_paging.py
cdm-ium: oh..... Interesting. I feel like this might break things in places. Does that mean this will start to fail in unit tests too? Do we need to bump marshmallow in `marquez` or something?

microcosm_flask/tests/conventions/fixtures.py
GabeLambda: As far as I, I think these are just for unit tests. I don't think we have builtin params that support this.

microcosm_flask/conventions/crud.py
GabeLambda: Right. I tried taking a look at the commit that introduced this, but didn't see anything interesting here. There may be a commit with the exact reasoning.

microcosm_flask/tests/test_paging.py
GabeLambda: That's a good question. The thing is that I think we actually do want errors on invalid parameters. Think of how many bugs there are when you pass in a parameter you think should exist, but actually doesn't.

GabeLambda:
Some other changes for downstream:
* change `pre/post_load` functions to accept kwargs ([ref](https://marshmallow.readthedocs.io/en/stable/upgrading.html#decorated-methods-and-handle-error-receive-many-and-partial)

afallou:
@gabe.lam since this introduces some breaking changes, I suggest bumping to a new major version.

microcosm_flask/conventions/crud.py
afallou: I dug down to the [original commit/PR](https://github.com/globality-corp/microcosm-flask/pull/7/files#diff-d021575c22812976dcde670b51504493R145) for this file and I couldn't find any comment related to this.

GabeLambda:
@afallou the next release will be a major version bump.

microcosm_flask/conventions/crud.py
afallou: I think that maybe Sphinx uses that ability to pass params in batch operations.

GabeLambda: I'll double-check this.

microcosm_flask/tests/test_paging.py
afallou: I agree that this is a good thing.

PR #236, opened 2019-09-09
Handle empty string for QueryStringList

afallou:
- Handle the case where an empty query param is passed for a `QueryStringList` field.
- Fix test for authentication error.
- Use `has_entries` instead of `equal_to`, to get more precise messages on test failure.

### DIFF
--- a/microcosm_flask/fields/query_string_list.py
+++ b/microcosm_flask/fields/query_string_list.py
@@ -18,17 +18,19 @@ class QueryStringList(List):
 
         /foo?bars=1,2
         /foo?bars[]=1&bars[]2
+        /foo?bars=&...
 
         and returns a list of values
 
         """
         if value is None:
             return None
+        if value == "":
+            return []
 
         try:
             attribute_elements = [attr_element.split(",") for attr_element in obj.getlist(attr)]
             attribute_params = [param for attr_param in attribute_elements for param in attr_param]
-
-            return PrintableList(super(QueryStringList, self)._deserialize(attribute_params, attr, obj))
+            return PrintableList(super()._deserialize(attribute_params, attr, obj))
         except ValueError:
             raise ValidationError("Invalid query string list argument")
--- a/microcosm_flask/tests/fields/test_query_string_list.py
+++ b/microcosm_flask/tests/fields/test_query_string_list.py
@@ -16,6 +16,10 @@ class QueryStringListSchema(Schema):
     foo_ids = QueryStringList(String())
 
 
+class NullableQueryStringListSchema(Schema):
+    foo_ids = QueryStringList(String(), allow_none=True)
+
+
 class TestEnum(Enum):
     A = "A"
     B = "B"
@@ -36,7 +40,7 @@ def test_query_list_deserialize_items():
 
 def test_query_list_load_with_comma_separated_single_keys():
     """
-    tests for support of /foo?foo_ids=1,2
+    tests for support of /foo?foo_ids=a,b
     """
     schema = QueryStringListSchema()
     result = schema.load(
@@ -48,7 +52,7 @@ def test_query_list_load_with_comma_separated_single_keys():
 
 def test_query_list_load_with_duplicate_keys():
     """
-    tests for support of /foo?foo_ids[]=1&foo_ids[]=2
+    tests for support of /foo?foo_ids[]=a&foo_ids[]=b
     """
     schema = QueryStringListSchema()
     result = schema.load(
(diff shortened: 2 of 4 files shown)


microcosm_flask/fields/query_string_list.py
```diff
@@ -18,17 +18,19 @@ def _deserialize(self, value, attr, obj, **kwargs):
...
             return None
+        if value == "":
+            return []
 
         try:
             attribute_elements = [attr_element.split(",") for attr_element in obj.getlist(attr)]
             attribute_params = [param for attr_param in attribute_elements for param in attr_param]
-
-            return PrintableList(super(QueryStringList, self)._deserialize(attribute_params, attr, obj))
+            return PrintableList(super()._deserialize(attribute_params, attr, obj))
```
afallou: We dropped support for Python 2.

microcosm_flask/fields/query_string_list.py
```diff
@@ -18,17 +18,19 @@ def _deserialize(self, value, attr, obj, **kwargs):
...
         /foo?bars[]=1&bars[]2
+        /foo?bars=&...
 
         and returns a list of values
 
         """
         if value is None:
             return None
+        if value == "":
+            return []
```
afallou: This is the main change.

microcosm_flask/tests/fields/test_query_string_list.py
```diff
@@ -58,6 +62,30 @@ def test_query_list_load_with_duplicate_keys():
...
+def test_none_query_list_load():
+    """
+    tests for support of /foo?foo_ids=&bar=...
+    """
+    schema = NullableQueryStringListSchema()
+    result = schema.load(
+        ImmutableMultiDict([("foo_ids", None)]),
+    )
+
+    assert_that(result["foo_ids"], is_(equal_to(None)))
```
afallou: Missing test for existing case.

microcosm_flask/tests/fields/test_query_string_list.py
```diff
@@ -58,6 +62,30 @@ def test_query_list_load_with_duplicate_keys():
...
+def test_empty_query_list_load():
+    """
+    tests for support of /foo?foo_ids=&bar=...
+    """
+    schema = QueryStringListSchema()
+    result = schema.load(
+        ImmutableMultiDict([("foo_ids", "")]),
+    )
+
+    assert_that(result["foo_ids"], is_(equal_to([])))
```
afallou: New test for emptry string case.

microcosm_flask/tests/test_basic_auth.py
```diff
@@ -32,14 +32,14 @@ def unauthorized():
...
-        "code": 401,
-        "message": "The server could not verify that you are authorized to access the URL requested. "
-                   "You either supplied the wrong credentials (e.g. a bad password), or your browser "
-                   "doesn't understand how to supply the credentials required.",
-        "retryable": False,
-        "context": {"errors": []},
-    })))
+    assert_that(data, has_entries(dict(
+        code=401,
+        message="The server could not verify that you are authorized to access the URL requested.  "
```
afallou: The missing space at the end of the line was making the test fail; and using `equal_to` instead `has_entries` makes it slower to pinpoint the issue.

microcosm_flask/tests/test_basic_auth.py
```diff
@@ -32,14 +38,12 @@ def unauthorized():
...
-        "code": 401,
-        "message": "The server could not verify that you are authorized to access the URL requested. "
-                   "You either supplied the wrong credentials (e.g. a bad password), or your browser "
-                   "doesn't understand how to supply the credentials required.",
-        "retryable": False,
-        "context": {"errors": []},
-    })))
+    assert_that(data, has_entries(dict(
+        code=401,
+        message=starts_with("The server could not verify that you are authorized to access the URL requested"),
```
afallou: The exact message has a different number of spaces locally vs. on CI. Testing for the whole message is not particularly interesting. Using `starts_with` is enough to make sure we report the error message.

microcosm_flask/tests/test_basic_auth.py
```diff
@@ -32,14 +38,12 @@ def unauthorized():
...
     data = loads(response.get_data().decode("utf-8"))
-    assert_that(data, is_(equal_to({
-        "code": 401,
-        "message": "The server could not verify that you are authorized to access the URL requested. "
-                   "You either supplied the wrong credentials (e.g. a bad password), or your browser "
-                   "doesn't understand how to supply the credentials required.",
-        "retryable": False,
-        "context": {"errors": []},
-    })))
+    assert_that(data, has_entries(dict(
```
afallou: Using `equal_to` makes it harder to track down what the issue is. We should use `has_entries` instead.

microcosm_flask/fields/query_string_list.py
mbaczynski: This will break the `value == str(deserialize(value))` symmetry. Not sure if this is a problem.

afallou: In the case of a query string, I don't think there's expectation of such a symmetry. One example is that a query string list can be passed as `foo[]=a&foo[]=b`

PR #237, opened 2019-09-09
Release version 2.0.3

afallou:
Handles empty string value for QueryStringList

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 2.0.2
+current_version = 2.0.3
 commit = False
 tag = False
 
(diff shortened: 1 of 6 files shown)


GabeLambda:
Do we actually want to support this? I would think this should be an error (e.g. like batch input, we should enforce min-1 entries).

afallou:
Discussed with @GabeLambda off-line. My take is:
- In a querystring we can't distinguish between `None` and `""`. Apparently werkzeug interprets it as `""`
- If we send `foo=[]` in a swagger client, we'll get `foo=&...` in the request.
- All in all, it means that we should interpret that empty string as an empty list. We also need validation to have a min-length of 1 for most of our `QueryStringList` field, but that's at the resource schema level.

GabeLambda:
I think it's worth mentioning that this essentially preserves status quo; we were doing this implicitly before the marshmallow 3.0 upgrade. This just avoids the validationErrors on empty input.

PR #238, opened 2019-10-09
Add support for enum formatting

GabeLambda:
This will serve as a signal for downstream clients to know (during
request construction/validation) how to properly format enums.

Note that there's no logic here directly; this merely adds the `format` key to the swagger definition.

Further note: `format` isn't part of the OpenAPI spec for enums specifically: https://swagger.io/docs/specification/data-models/data-types/. It will be added as part of another (internal) PR.


GabeLambda:
Done.

PR #240, opened 2019-10-20
Introduce associated schemas

afallou:
Currently, schema definitions that we expose in the swagger convention are all tied to a route.
Here we introduce the concept of an "associated schema", which is derived from another resource but not tied to a route.

One use case we have is for pubsub: our pubsub messages are closely tied to our route resources; however we don't necessarily send all of a resource fields in a pubsub message. The mechanism introduced here will allow us to use swagger schemas to define those message schemas.

- Introduce `add_associated_schema` decorator.
- Use a schema's associated schemas when generating swagger definitions.

### DIFF
--- a/microcosm_flask/decorators/schemas.py
+++ b/microcosm_flask/decorators/schemas.py
@@ -1,4 +1,5 @@
 from copy import deepcopy
+from dataclasses import dataclass
 
 from inflection import camelize, underscore
 from marshmallow import Schema
@@ -7,12 +8,20 @@ from microcosm_flask.naming import name_for
 from microcosm_flask.swagger.naming import type_name
 
 
+@dataclass
+class SelectedField:
+    name: str
+    required: bool = True
+
+
 def _get_fields_and_make_required(schema_cls, selected_fields):
     associated_fields = {}
-    for field_name in selected_fields:
-        field = deepcopy(schema_cls._declared_fields[field_name])
-        field.required = True
-        associated_fields[field_name] = field
+    for selected_field in selected_fields:
+        if isinstance(selected_field, str):
+            selected_field = SelectedField(selected_field)
+        schema_field = deepcopy(schema_cls._declared_fields[selected_field.name])
+        schema_field.required = selected_field.required
+        associated_fields[selected_field.name] = schema_field
 
     return associated_fields
 
@@ -29,9 +38,13 @@ def add_associated_schema(name_suffix, selected_fields=()):
     """
     Derive a schema as a subset of fields from the schema class being decorated,
     and add that derived schema as an attribute on the decorated schema.
-    All fields from the derived schema are marked as required.
 
     This allows us to expose the derived schema in the swagger definition.
+
+    :name_suffix: Suffix that will added to the decorator schema's definition name to give the
+                  name of the exposed definition
+    :selected_fields: List of `SelectedField` instances. As a shorthand a list of strings can also
+                      be passed, which will make all sected field required
 
     """
     def decorator(schema_cls):
--- a/microcosm_flask/decorators/schemas.py
+++ b/microcosm_flask/decorators/schemas.py
@@ -1,0 +1,57 @@
+from copy import deepcopy
+
+from inflection import camelize, underscore
+from marshmallow import Schema
+
+from microcosm_flask.naming import name_for
+from microcosm_flask.swagger.naming import type_name
+
+
+def _get_fields_and_make_required(schema_cls, selected_fields):
+    associated_fields = {}
+    for field_name in selected_fields:
+        field = deepcopy(schema_cls._declared_fields[field_name])
+        field.required = True
+        associated_fields[field_name] = field
+
+    return associated_fields
+
+
+def associated_schemas_attr_name(schema_cls):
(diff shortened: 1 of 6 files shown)


microcosm_flask/tests/conventions/fixtures.py
```diff
@@ -81,6 +83,13 @@ def csv_column_order(self):
...
 
 
+@add_associated_schema(
+    "Foo",
+    [
+        "email",
+        "firstName",
+    ]
+)
 class PersonSchema(PersonCSVSchema):
```
afallou: Example use of this new decorator. I added the `email` field in the `NewPerson` schema to have a non-required field.

microcosm_flask/tests/conventions/test_crud.py
```diff
@@ -159,7 +159,7 @@ def test_search_with_context(self):
         })
 
     def test_reuse_search_self_link(self):
-        uri = "/api/address?list_param=a,b,c&enum_param=A".format(PERSON_ID_1)
```
afallou: Unrelated changes, but those `format` were unnecessary, and mypy has gotten stricter and was erroring on them.

microcosm_flask/tests/swagger/test_definitions.py
```diff
@@ -44,12 +44,12 @@ def test_build_swagger():
         operations = list(iter_endpoints(graph, match_function))
         swagger_schema = build_swagger(graph, ns, operations)
 
-    assert_that(swagger_schema, is_(equal_to({
```
afallou: I changed the straight `equal_to` to nested `has_entries` to make finding issues on test failure easier.

microcosm_flask/tests/swagger/test_definitions.py
```diff
@@ -168,47 +176,63 @@ def test_build_swagger():
...
+                    },
+                    "firstName": {
+                        "type": "string",
+                    },
+                },
+                required=[
+                    "email",
+                    "firstName",
+                ],
+            ),
```
afallou: This is a new definition introduced by using the new decorator.

microcosm_flask/tests/conventions/fixtures.py
```diff
@@ -81,6 +83,13 @@ def csv_column_order(self):
         return column_order
 
 
+@add_associated_schema(
+    "Foo",
```
cdm-ium: I'd give this a meaningful name otherwise the intent is confusing (I don't understand what it is). Is it a "PersonSubSchema" or a "PersonPubsubSchema"?

cdm-ium: Maybe associatred schemas could use an extra property to specify their intent? We are basically adding a feature for the webservice to expose information about schemas that it may not use, so we may want to create a facility for those schemas to document themeslves a little more.

afallou: Good points. I'll update the tests and see if some "docstring" can be exposed in swagger.

microcosm_flask/tests/conventions/test_crud.py
GabeLambda: I think we actually also have a better method of search here for query params (as an arg to the `get()` call below).

microcosm_flask/decorators/schemas.py
```diff
@@ -0,0 +1,57 @@
...
+
+from microcosm_flask.naming import name_for
+from microcosm_flask.swagger.naming import type_name
+
+
+def _get_fields_and_make_required(schema_cls, selected_fields):
+    associated_fields = {}
+    for field_name in selected_fields:
+        field = deepcopy(schema_cls._declared_fields[field_name])
+        field.required = True
```
GabeLambda: We don't want to be able to specify non-required fields? I know we may not have a use case, but others might.

microcosm_flask/tests/conventions/fixtures.py
```diff
@@ -81,6 +83,19 @@ def csv_column_order(self):
         return column_order
 
 
+def pubsub_schema(fields):
```
GabeLambda: Is there a reason you're not making this part of the public API? I understand if the intent is to reduce coupling of terms, but it feels like this is the main usage, and we'd be copy-pasting it around. Unless your intent is more that we redefine the convention on every project, which I'm not sure I agree with.

afallou: It's because I don't think `microcosm-flask` should know/deal with pubsub or any other use case of this; all it's doing is exposing extra schemas. That's why I went with a fairly generic API.
I was going to define this one in `marquez`

microcosm_flask/decorators/schemas.py
afallou: I hesitated, because making all fields required just simplifies the API. I guess I could leave the option specify optional field, and make a shorthand version that makes all fields required.

microcosm_flask/tests/conventions/test_crud.py
afallou: We do. You caught me being lazy.

microcosm_flask/tests/conventions/fixtures.py
GabeLambda: 👍

PR #246, opened 2019-11-04
Allow different parent classes for associated schema

afallou:
Allow overriding the base class when defining an associated schema.

### DIFF
--- a/microcosm_flask/decorators/schemas.py
+++ b/microcosm_flask/decorators/schemas.py
@@ -14,7 +14,7 @@ class SelectedField:
     required: bool = True
 
 
-def _get_fields_and_make_required(schema_cls, selected_fields):
+def _get_fields_from_schema(schema_cls, selected_fields):
     associated_fields = {}
     for selected_field in selected_fields:
         if isinstance(selected_field, str):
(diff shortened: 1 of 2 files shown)


microcosm_flask/decorators/schemas.py
```diff
@@ -14,7 +14,7 @@ class SelectedField:
     required: bool = True
 
 
-def _get_fields_and_make_required(schema_cls, selected_fields):
+def _get_fields_from_schema(schema_cls, selected_fields):
```
afallou: This name was wrong since fields can be optional.

cdm-ium:
One thing here for posterity---why is this feature useful? what do you use it for?

At first glance I'm a little confused for the usage as it looks like this allows you to create a schema based off of a decorated schema, but to use a totally separate base class.

afallou:
@cdm-ium that's a fair point.
The decorator allows us to expose a new schema in the openAPI spec, based on an existing schema.
This inheritance mechanism is just a way of getting some flexibility, allowing us to have some fields that are common to all exposed schemas.

It's really meant to be used by wrapping this decorator inside another one. Our example is pubsub where:
- We want to expose to have pubsub message schemas derived from resource schemas (the decorator allows us to do that).
- We also want those messages schemas to all have `uri` and `mediaType` fields.

PR #248, opened 2019-11-14
Expose function to set an associated schema

afallou:
- Adding an associated schema can be useful outside of the context of the `@add_associated_schema` decorator; separate that function out so as to expose it.
- Merge `master` back to `develop`.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 2.2.0
+current_version = 2.4.0
 commit = False
 tag = False
 
--- a/microcosm_flask/decorators/schemas.py
+++ b/microcosm_flask/decorators/schemas.py
@@ -41,6 +41,27 @@ def get_associated_schema(schema_cls, name_suffix):
     raise KeyError(f"Schema {schema_cls} does not have an associated schema with suffix {name_suffix}")
 
 
+def set_associated_schema(target_cls, name_suffix, associated_schema):
+    """
+    Add a derived schema as an attribute to a schema class, using conventions
+    that will result in the associated schema being exposed in the OpenAPI definition of the service.
+
+    """
+    # Use the class name in the attribute name to avoid sharing with children classes
+    attr_name = associated_schemas_attr_name(target_cls)
+    try:
+        associated_schemas = getattr(target_cls, attr_name)
+        if name_suffix in associated_schemas:
+            raise ValueError(f"Schema {target_cls} already has an associated schema for suffix {name_suffix}")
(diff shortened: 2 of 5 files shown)


.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 2.2.0
+current_version = 2.4.0
```
afallou: This is from merging back `master`.

microcosm_flask/decorators/schemas.py
```diff
@@ -41,6 +41,27 @@ def get_associated_schema(schema_cls, name_suffix):
...
+        associated_schemas = getattr(target_cls, attr_name)
+        if name_suffix in associated_schemas:
+            raise ValueError(f"Schema {target_cls} already has an associated schema for suffix {name_suffix}")
+        associated_schemas[name_suffix] = associated_schema
+    except AttributeError:
+        setattr(
+            target_cls,
+            attr_name,
+            {name_suffix: associated_schema},
+        )
```
globality-thill: Do we need to have a test for this now it is a function of its own?

microcosm_flask/decorators/schemas.py
```diff
@@ -41,6 +41,35 @@ def get_associated_schema(schema_cls, name_suffix):
...
+
+    """
+    # Use the class name in the attribute name to avoid sharing with children classes
+    attr_name = associated_schemas_attr_name(target_cls)
+    try:
+        associated_schemas = getattr(target_cls, attr_name)
+        if name_suffix in associated_schemas:
+            raise ValueError(f"Schema {target_cls} already has an associated schema for suffix {name_suffix}")
+        associated_schemas[name_suffix] = associated_schema
+    except AttributeError:
```
rh-glob: Can this be tighter in scope only encompassing `associated_schemas = getattr(target_cls, attr_name)`? As that should be only line we expect an `AttributeError` from.

microcosm_flask/decorators/schemas.py
rh-glob: make it private/protected and problem solved? 🤣

afallou: We test it indirectly via existing tests, but you're right. Added a unit test specifically for this one.

microcosm_flask/decorators/schemas.py
afallou: Reproducing IRL conversation here:
- There isn't really a way to tighten the scope while using `try/catch`
- But I think that's acceptable, situations like that are a bit of a side effect of using exceptions for control flow.

PR #254, opened 2020-01-09
GLOB-36585 set query parameters to be required based on their schema

sammy-globality:
Currently query parameters are always set as having `required=False` when building the operation for swagger. Either I'm missing some crucial understanding about how we use swagger, or this has been an unnoticed bug for 4 years.

I've changed it to be based on the field specification and updated the one test that checks this.

### DIFF
--- a/microcosm_flask/tests/conventions/test_query.py
+++ b/microcosm_flask/tests/conventions/test_query.py
@@ -120,7 +120,7 @@ class TestQuery:
                             "type": "string",
                         },
                         {
-                            "required": False,
+                            "required": True,
                             "type": "string",
                             "name": "value",
                             "in": "query",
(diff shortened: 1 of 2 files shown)


microcosm_flask/tests/conventions/test_query.py
```diff
@@ -120,7 +120,7 @@ def test_swagger(self):
                             "type": "string",
                         },
                         {
-                            "required": False,
+                            "required": True,
```
GabeLambda: Add another case here to make sure that:
* non-required fields are marked as such
* required fields are similarly marked

sammy-globality:
> So long as you address @GabeLambda 's comments around tests.

Already did :)

PR #256, opened 2020-01-13
Add hostname and IP to flask

KensoDev:
We want to see the actual hostname and IP of the server returning the call.
This isn't a security risk *for us* due to the fact that the health checks are not exposed to the world.

Since this is an open-source project, we might wrap this with a flag in the future.

### DIFF
--- a/microcosm_flask/conventions/health.py
+++ b/microcosm_flask/conventions/health.py
@@ -6,6 +6,7 @@ using HTTP 200/503 status codes to indicate healthiness.
 
 """
 from distutils.util import strtobool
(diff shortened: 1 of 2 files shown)


KensoDev:
Yes, this could also be added as a custom check, just like PG or ES/Memcache checks we do.

KensoDev:
Thanks @mlcohen . I'll wait for other people to 👁 before we merge this in.

GabeLambda:
If we're making a note that this is a potential security leak now, we can add it as a flag here. Especially since we might be using this as a debugging mechanism, rather than something we'd want always on by default.

PR #257, opened 2020-01-19
Fix building OpenAPI definition for integer-valued ids

afallou:
**Why?**
The way we build OpenAPI definitions for flask routes breaks on non-string valued ids.
In particular, it breaks for integer-valued ids. This is because we use werkzeug functions to build paths in the OpenAPI definitions; using parameter names instead of actual values. This works fine for string-valued parameters, but not in other cases.
In particular, it breaks for integer-valued IDs.

**What?**
Use placeholder integers to build URIs using werkzeug functions, and replace placeholders with param names afterwards.

### DIFF
--- a/microcosm_flask/swagger/definitions.py
+++ b/microcosm_flask/swagger/definitions.py
@@ -17,6 +17,7 @@ Note that:
 
 """
 from logging import getLogger
+from typing import Set
 from urllib.parse import unquote
 
 from openapi import model as swagger
@@ -35,6 +36,8 @@ from microcosm_flask.swagger.naming import operation_name, type_name
 
 
 logger = getLogger("microcosm_flask.swagger")
+# Placeholder integer id used to build URIs in werkzeug before replacing with id param name.
+# Use a value that is unlikely to be present somewhere else in the URI.
 PLACEHOLDER_INTEGER_ID = 1111
 
 
@@ -121,7 +124,7 @@ def iter_definitions(definitions, operations):
         yield get_response_schema(func)
 
 
-def build_path_for_integer_param(ns, operation, arguments):
+def build_path_for_integer_param(ns, operation, arguments: Set):
     """
     Build an endpoint path when the parameters are integer-valued
 
@@ -138,10 +141,10 @@ def build_path_for_integer_param(ns, operation, arguments):
     them afterwards.
 
     """
-    sorted_args = list(arguments)
+    ordered_args = list(arguments)
     args_substitution = {
         PLACEHOLDER_INTEGER_ID + index: argument
-        for index, argument in enumerate(sorted_args)
+        for index, argument in enumerate(ordered_args)
     }
     uri_templates = {
         argument: f"{placeholder}"
--- a/microcosm_flask/swagger/definitions.py
+++ b/microcosm_flask/swagger/definitions.py
@@ -35,6 +35,7 @@ from microcosm_flask.swagger.naming import operation_name, type_name
 
 
 logger = getLogger("microcosm_flask.swagger")
+PLACEHOLDER_INTEGER_ID = 1111
 
 
 def build_swagger(graph, ns, operations):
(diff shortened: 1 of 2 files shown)


microcosm_flask/swagger/definitions.py
```diff
@@ -120,21 +121,53 @@ def iter_definitions(definitions, operations):
...
+    That works with UUID-valued ids but not integer-valued ones, due do a difference
+    between the UUIDConverter and NumberConverter's `to_url` methods
+    (see https://github.com/pallets/werkzeug/blob/master/src/werkzeug/routing.py#L1315
+    and https://github.com/pallets/werkzeug/blob/master/src/werkzeug/routing.py#L1234)
+
+    Here we instead use placeholder integers to use werkzeug functions, and replace
+    them afterwards.
 
     """
+    sorted_args = list(arguments)
```
rh-glob: How does casting them to a list making them sorted? are you missing a `sorted` or is the input to this function expecting this to be pre-sorted?

microcosm_flask/swagger/definitions.py
```diff
@@ -120,21 +121,53 @@ def iter_definitions(definitions, operations):
         yield get_response_schema(func)
 
 
-def build_path(operation, ns):
+def build_path_for_integer_param(ns, operation, arguments):
```
rh-glob: A test of this directly would be useful, especially to see the structure of the outputted value

microcosm_flask/swagger/definitions.py
```diff
@@ -35,6 +35,7 @@
 
 
 logger = getLogger("microcosm_flask.swagger")
+PLACEHOLDER_INTEGER_ID = 1111
```
rh-glob: Any reasons for this particular value or does it just need to be suitably large?

microcosm_flask/swagger/definitions.py
afallou: `arguments` is a set. `ordered_args` would be a better name than `sorted_args`.

microcosm_flask/swagger/definitions.py
afallou: I wanted something that wouldn't already be in the URI by chance.

microcosm_flask/swagger/definitions.py
rh-glob: Do we using `typing` in this library? If so might be nice to add the type definitions to help

microcosm_flask/swagger/definitions.py
```diff
@@ -120,21 +121,53 @@ def iter_definitions(definitions, operations):
...
     except BuildError as error:
-        # we are missing some URI path parameters
-        uri_templates = {
-            argument: "{{{}}}".format(argument)
-            for argument in error.suggested.arguments
-        }
-        # flask will sometimes try to quote '{' and '}' characters
-        return unquote(ns.url_for(operation, _external=False, **uri_templates))
+        if ns.identifier_type == "int":
+            return build_path_for_integer_param(ns, operation, error.suggested.arguments)
```
rh-glob: Can this be achieved before the error is thrown or is the only place to get the arguments? Also does the `error.suggested.arguments` contain all the arguments or just the ones that had issues? If the latter are non-problematic arguments correctly handled?

afallou: `error.suggested.arguments` only contains the params that had issues. In this context however, all param will error because we're not passing in values for any of them.

The idea is that we're asking werkzeug to build a URL like `/api/v1/person/:person_id` but we don't have a `person_id`. So we let werkzeug tell us what's missing, and put the param name in there (e.g. `{person_id}`); and we end up with the right format for OpenAPI.

rh-glob: Does `microcosm_flask` support  building urls such as: `/api/v1/users/<uuid>/company/<int>/queue/<int>/`? As it looks like these will fail to build correctly as code relies on no nesting of various types in a single url.

cdm-ium: yeah i don't think we do doulbe uuid urls.

GabeLambda: Right, we mostly just go one layer deep (e.g. `/api/v1/company/<id>/event`), when we do use that convention.

PR #259, opened 2020-01-29
Add `x-nullable` for nullable fields.

afallou:
**Why?**
OpenAPI 2.0 doesn't support nullable fields natively, so we had taken the approach to mark those fields as non-required.
However, that approach doesn't cover all cases. Bravado supports the custom `x-nullable` flag for a field.

**What?**
Mark fields as `x-nullable` in the OpenAPI definition if they have `allow_none` set to `True` in their schema.

### DIFF
--- a/microcosm_flask/tests/swagger/test_definitions.py
+++ b/microcosm_flask/tests/swagger/test_definitions.py
@@ -166,6 +166,7 @@ def test_build_swagger():
                     "email": {
                         "type": "string",
                         "format": "email",
+                        "x-nullable": True,
                     },
                     "lastName": {
                         "type": "string",
@@ -188,6 +189,7 @@ def test_build_swagger():
                     "email": {
                         "format": "email",
                         "type": "string",
+                        "x-nullable": True,
                     },
                     "firstName": {
                         "type": "string",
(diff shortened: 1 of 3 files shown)


afallou:
Note: it seems like this was already done in https://github.com/globality-corp/microcosm-flask/pull/85 but reverted due to https://github.com/Yelp/bravado-core/pull/143.
Now that the bravado-core bug is fixed we should be good.

microcosm_flask/tests/swagger/test_definitions.py
```diff
@@ -166,6 +166,7 @@ def test_build_swagger():
                     "email": {
                         "type": "string",
                         "format": "email",
+                        "x-nullable": True,
```
cdm-ium: Does this mean we have to regenerate all of our swagger schemas so this gets generated into them or is this already present in our schemas?

afallou: We will need to regenerate.

afallou:
So https://github.com/globality-corp/microcosm-flask/pull/85 is the more complete way of doing this change, but it's also not backwards-compatible, so I won't go down that route.

PR #261, opened 2020-01-29
Release 2.10.0

afallou:
Releases https://github.com/globality-corp/microcosm-flask/pull/259 – see there for more context.


.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 2.8.0
+current_version = 2.10.0
```
afallou: This will make more sense once https://github.com/globality-corp/microcosm-flask/pull/260 is merged.

PR #264, opened 2020-03-24
Support data_key usage for field names for generated swagger

GabeLambda:
The recommended method of specifying serialized names in marshmallow is
to use the `data_key` argument when specifying a field, instead of `attribute`
(which otherwise leads to the awkward inclusion of camelCased field names,
which is less than ideal).

Unfortunately, this otherwise breaks existing swagger definition
generation, as it didn't take this field into account.

This will use an existing `data_key` field if it exists as a name, but
otherwise fallback to the predefined name.

https://marshmallow.readthedocs.io/en/stable/quickstart.html#specifying-serialization-deserialization-keys

### DIFF
--- a/microcosm_flask/swagger/schemas.py
+++ b/microcosm_flask/swagger/schemas.py
@@ -38,7 +38,7 @@ class Schemas:
         fields = list(self.iter_fields(schema))
 
         properties = {
-            field.name: self.build_parameter(field)
+            name: self.build_parameter(field)
             for name, field in fields
         }
 
@@ -67,7 +67,9 @@ class Schemas:
         """
         for name in sorted(schema.fields.keys()):
             field = schema.fields[name]
-            yield name, field
+            field_name = field.data_key or name
+
+            yield field_name, field
 
     def iter_schemas(self, schema: Schema) -> Iterable[Tuple[str, Any]]:
         """
--- a/microcosm_flask/tests/conventions/fixtures.py
+++ b/microcosm_flask/tests/conventions/fixtures.py
@@ -28,12 +28,13 @@ class Person:
 
 
 class NewAddressSchema(Schema):
-    addressLine = fields.Str(attribute="address_line", required=True)
+    addressLine = fields.String(attribute="address_line", required=True)
 
 
 class NewPersonSchema(Schema):
-    firstName = fields.Str(attribute="first_name", required=True)
-    lastName = fields.Str(attribute="last_name", required=True)
+    # both attribute and data_key fields should result in the same swagger definition
+    firstName = fields.String(attribute="first_name", required=True)
+    last_name = fields.String(data_key="lastName", required=True)
     email = fields.Email()
 
     @property
@@ -46,8 +47,8 @@ class NewPersonBatchSchema(Schema):
 
 
 class UpdatePersonSchema(Schema):
-    firstName = fields.Str(attribute="first_name")
-    lastName = fields.Str(attribute="last_name")
+    firstName = fields.String(attribute="first_name")
+    last_name = fields.String(data_key="lastName")
 
 
 class AddressCSVSchema(NewAddressSchema):


microcosm_flask/swagger/schemas.py
```diff
@@ -65,9 +65,15 @@ def iter_fields(self, schema: Schema) -> Iterable[Tuple[str, Field]]:
         Generates: name, field pairs
 
         """
+        serialized_names = {}
+        for field_name, field_def in schema.declared_fields.items():
+            if isinstance(field_def, Field) and field_def.data_key:
+                serialized_names[field_name] = field_def.data_key
```
GabeLambda: Full disclosure: I stole this straight from: https://www.programcreek.com/python/example/106766/marshmallow.fields.Field

microcosm_flask/swagger/schemas.py
```diff
@@ -65,9 +65,15 @@ def iter_fields(self, schema: Schema) -> Iterable[Tuple[str, Field]]:
         Generates: name, field pairs
 
         """
+        serialized_names = {}
+        for field_name, field_def in schema.declared_fields.items():
+            if isinstance(field_def, Field) and field_def.data_key:
```
afallou: What cases is this `isinstance(field_def, Field)` supposed to guard against?

microcosm_flask/swagger/schemas.py
```diff
@@ -65,9 +65,15 @@ def iter_fields(self, schema: Schema) -> Iterable[Tuple[str, Field]]:
...
+                serialized_names[field_name] = field_def.data_key
+
         for name in sorted(schema.fields.keys()):
```
afallou: It seems weird to be iterating once over `schema.declared_fields` and once over `schema.fields`. Seems like we could have a single loop.

rh-glob: Removing the loop and the tests pass fine
```
diff --git a/microcosm_flask/swagger/schemas.py b/microcosm_flask/swagger/schemas.py
index d1a8d87..1bd27a9 100644
--- a/microcosm_flask/swagger/schemas.py
+++ b/microcosm_flask/swagger/schemas.py
@@ -65,15 +65,11 @@ class Schemas:
         Generates: name, field pairs

         """
-        serialized_names = {}
-        for field_name, field_def in schema.declared_fields.items():
-            if isinstance(field_def, Field) and field_def.data_key:
-                serialized_names[field_name] = field_def.data_key
-
         for name in sorted(schema.fields.keys()):
             field = schema.fields[name]
-            serialized_name = serialized_names.get(name, name)
-            yield serialized_name, field
+            if isinstance(field, Field) and field.data_key:
+                name = field.data_key
+            yield name, field

     def iter_schemas(self, schema: Schema) -> Iterable[Tuple[str, Any]]:
         """
```

Is there some other case you a guarding against with the separate loop @GabeLambda ?

microcosm_flask/swagger/schemas.py
GabeLambda: I'm guessing this is mostly a belts-and-suspenders type deal.  I can remove it, it's not adding clarity.

microcosm_flask/swagger/schemas.py
GabeLambda: From memory, there was some discrepancy I had found between fields and declared_fields (one had `data_key`, the other didn't). But apparently I was mistaken.

microcosm_flask/tests/conventions/fixtures.py
```diff
@@ -46,8 +47,8 @@ class NewPersonBatchSchema(Schema):
 
 
 class UpdatePersonSchema(Schema):
-    firstName = fields.Str(attribute="first_name")
-    lastName = fields.Str(attribute="last_name")
+    firstName = fields.String(attribute="first_name")
+    last_name = fields.String(data_key="lastName")
```
afallou, line 51: I'm guessing the test assertions don't need to be updated because the output is exactly the same?

GabeLambda: Indeed. The end swagger shouldn't have any changes.

PR #268, opened 2020-04-21
Add DeleteBatch operation

dumpstate:
Add `DeleteBatch` operation to allow endpoints like:

```
DELETE /collection
```

with optional query params, returning `204` on success with empty body.

### DIFF
--- a/microcosm_flask/conventions/crud.py
+++ b/microcosm_flask/conventions/crud.py
@@ -195,8 +195,8 @@ class CRUDConvention(Convention):
             definition.header_func(headers, response_data)
             response_format = self.negotiate_response_content(definition.response_formats)
             return dump_response_data(
-                "",
-                None,
+                response_schema="",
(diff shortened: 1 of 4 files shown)


microcosm_flask/conventions/crud.py
```diff
@@ -170,6 +170,40 @@ def update_batch(**path_data):
...
+        @qs(request_schema)
+        @wraps(definition.func)
+        def delete_batch(**path_data):
+            headers = dict()
+            request_data = load_query_string_data(request_schema)
+            response_data = require_response_data(definition.func(**merge_data(path_data, request_data)))
+            definition.header_func(headers, response_data)
+            response_format = self.negotiate_response_content(definition.response_formats)
+            return dump_response_data(
+                "",
```
rh-glob, line 198: Can you add the named kwargs for clarity on what values these are setting

PR #272, opened 2020-04-22
GLOB-44246 Improve swagger endpoint performance

GabeLambda:
As the number of endpoints grows, the time taken for this endpoint grows 
non-linearly. This is due to the fact that during `add_paths`, `url_for` 
performs what is essentially a full scan across every possible endpoint 
that the app is already aware of, performing a relatively expensive diff on
each turn.

While this only occurs on BuildError, it is frequently raised when the
endpoint in question requires any path argument, which ends up being many
of them
(retrieve/delete/update/etc.).

We can sidestep that error by performing a preflight operation (which
`url_for` is already end up doing under the hood) that pre-fetches the 
expected arguments used in any given path, and passes that in when 
constructing the path.

Also add an extra flag to enable schema validation, as this can frequently
eat up a non-trivial amount of time. This will, however, push the burden of
validating schemas onto the project itself (e.g. via an additional unit
test).

From local testing (with a schema using 200+ endpoints) this reduced the total time from ~6 seconds to ~1.

### DIFF
--- a/microcosm_flask/swagger/definitions.py
+++ b/microcosm_flask/swagger/definitions.py
@@ -17,11 +17,18 @@ Note that:
 
 """
 from logging import getLogger
-from typing import Set
+from typing import (
+    Dict,
+    List,
+    Optional,
+    Set,
+    Tuple,
+)
 from urllib.parse import unquote
 
+from flask.globals import _request_ctx_stack
 from openapi import model as swagger
-from werkzeug.routing import BuildError
+from werkzeug.routing import BuildError, Rule
 
 from microcosm_flask.conventions.registry import (
     get_qs_schema,
@@ -29,6 +36,7 @@ from microcosm_flask.conventions.registry import (
     get_response_schema,
 )
 from microcosm_flask.errors import ErrorContextSchema, ErrorSchema, SubErrorSchema
+from microcosm_flask.namespaces import Namespace
 from microcosm_flask.naming import name_for
 from microcosm_flask.operations import Operation
 from microcosm_flask.swagger.api import build_parameter, iter_schemas
@@ -41,7 +49,7 @@ logger = getLogger("microcosm_flask.swagger")
 PLACEHOLDER_INTEGER_ID = 1111
 
 
-def build_swagger(graph, ns, operations):
+def build_swagger(graph, ns, operations, validate_schema=False):
     """
     Build out the top-level swagger definition.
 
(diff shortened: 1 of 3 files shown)


microcosm_flask/swagger/definitions.py
```diff
@@ -65,22 +73,43 @@ def build_swagger(graph, ns, operations):
     )
     add_paths(schema.paths, base_path, operations)
     add_definitions(schema.definitions, operations)
-    try:
-        schema.validate()
-    except Exception:
-        logger.exception("Swagger definition did not validate against swagger schema")
-        raise
+
+    if validate_schema:
```
GabeLambda, line 77: This will mean any service using this lib will need to add another test to validate their specific schema.

microcosm_flask/swagger/definitions.py
```diff
@@ -65,22 +73,43 @@ def build_swagger(graph, ns, operations):
...
 
+def _construct_schema_request_arguments():
+    """
+    Create a mapping between a given endpoint/method and its possible arguments.
+    Used in aid of generating paths during swagger generation
+
+    """
+    reqctx = _request_ctx_stack.top
+    url_adapter = reqctx.url_adapter
+    rules: List[Rule] = url_adapter.map._rules
```
GabeLambda, line 95: This is essentially what `url_for` is referencing in that flow.

microcosm_flask/swagger/definitions.py
```diff
@@ -157,19 +186,34 @@ def build_path_for_integer_param(ns, operation, arguments: Set):
...
+    )
+
+    expected_uri_templates = create_uri_templates(expected_arguments)
+
     try:
-        return ns.url_for(operation, _external=False)
-    except BuildError as error:
+        # flask will sometimes try to quote '{' and '}' characters
+        return unquote(ns.url_for(operation, _external=False, **expected_uri_templates))
+    except (BuildError, ValueError) as error:
```
GabeLambda, line 208: This section mainly exists to cover any bases that may have been missed in the above check. It may be the case that ValueError (i.e. from the int-based part argument) is the only one that's really required, in which case, I can just remove the fallback and just handle that case here.

GabeLambda: Scratch that, there are some endpoints (e.g. alias, upload_for) where the pre-fetching strategy fails. I could dig into that more, or just allow the fallback to keep existing here.

PR #273, opened 2020-05-25
Add sentry support

rh-glob:
- Add support to configure a sentry client to report errors, alongside current json error responses
- sentry package is optional
- sentry setup/configuration is taken from `microcosm_pubsub` with some minor changes. This should be moved to a `microcosm_sentry` library at a later date. I have avoided that for the moment as want to get full tests with a service in place before that to see what functionality should live there and what needs to be configurable by the consumer of the library.


microcosm_flask/errors.py
```diff
@@ -147,14 +147,34 @@ def make_json_error(error):
...
+    try:
+        send_error_to_sentry(error, response)
+    except ImportError:
+        pass
+    return response
+
+
+def send_error_to_sentry(error, response):
+    from sentry_sdk import capture_exception
+    from sentry_sdk import configure_scope
```
afallou, line 161: I understand that those are here to avoid loading all the sentry stuff if we don't need them; but I think it highlights that having a separate component to send exceptions to sentry would make sense.

rh-glob: Agreed. I have a rough structure sketched out for a `microcosm_flask` library and this was an element to be included. Hesitant to include this component as part of this PR would rather wait for lib. Want to get the pubsub/flask parts enabled for a service to see what added metadata I need to add to make each function well in sentry.

As development will be even slower once I need to make this it's own library as to see any sort of change to a service I would need to:
- merge to sentry lib
- release that
- update flask/pubsub lib
- release that
- update service
Not exactly conducive so a iterative development cycle 😆

simpleigh-globality: Woo `microcosm` monorepo!

(I'll get my coat).

microcosm_flask/sentry.py
```diff
@@ -0,0 +1,67 @@
...
+                        continue
+                    elif var_name.endswith("_id"):
+                        continue
+                frame["vars"][var_name] = "<redacted>"
+    return event
+
+
+@defaults(
+    enabled=False,
+    dsn=None,
```
dumpstate, line 40: It took me a while to figure out what `dsn` stands for (very sentry specific), maybe a comment? Or accept full `data_source_name` in config?

microcosm_flask/sentry.py
```diff
@@ -0,0 +1,67 @@
...
+                sentry = sentry_sdk.init(
+                    graph.config.sentry_logging.dsn,
+                    server_name=f"{graph.metadata.name}_api",
+                    release=graph.config.build_info_convention.sha1,
+                    environment=environ.get("MICROCOSM_ENVIRONMENT") or "undefined",
+                    before_send=before_send,
+                )
+            except BadDsn:
+                enabled, dsn, sentry = False, None, None
+                configure_sentry.logger.error("Bad DSN value set")
```
dumpstate, line 62: what not to raise an exception?

microcosm_flask/sentry.py
```diff
@@ -0,0 +1,67 @@
...
+from microcosm_logging.decorators import logger
+
+from sentry_sdk.utils import BadDsn
+
+
+@dataclass
+class SentryConfig:
+    dsn: Optional[str] = None
+    enabled: bool = False
+    client: Optional[Any] = None
```
dumpstate, line 17: Can it be an actual client type from `sentry_sdk`?

microcosm_flask/sentry.py
rh-glob: As this is configured from outside that app I didn't want an outside variable potentially disrupting/breaking the app loading. So opted to handle the case where bad config is supplied and disable the component completely.

microcosm_flask/sentry.py
rh-glob: How the client is exposed in `sentry_sdk` is quite weird it goes through a few layers of lamdas which didn't play nice when I tried to add typing info

microcosm_flask/sentry.py
rh-glob: Good point will add a comment, as sorta locked into the config naming due to using this in the pubsub implementation. Will look into updating these when these get replaced by a lib (also probably worth adding a `_url` suffix too)

PR #279, opened 2020-06-02
Add option for non-strict enums in OpenAPI definition

afallou:
**Why?**
Having closed enums on the response validation side breaks backward-compatibility.

Let's say we have two services `consumer` and `producer`. `producer` gets a new version of a shared enum, with a new enum value.
If `producer` returns that value it in the body of a response, `consumer` will always error on deserializing that response even if that new enum value is backward-compatible: for example, `consumer` may not be using that enum at all.

**What?**
Allow enums to be output as simple strings/ints to the OpenAPI definition, making then effectively open-ended from the client validation perspective.

### DIFF
--- a/microcosm_flask/swagger/api.py
+++ b/microcosm_flask/swagger/api.py
@@ -6,6 +6,7 @@ from typing import (
     Any,
     Iterable,
     Mapping,
+    Optional,
     Tuple,
 )
 
@@ -16,30 +17,30 @@ from microcosm_flask.swagger.parameters import Parameters
 from microcosm_flask.swagger.schemas import Schemas
 
 
-def build_schema(schema: Schema) -> Mapping[str, Any]:
+def build_schema(schema: Schema, strict_enums: Optional[bool] = True) -> Mapping[str, Any]:
     """
     Build JSON schema from a marshmallow schema.
 
     """
-    builder = Schemas(build_parameter=build_parameter)
+    builder = Schemas(build_parameter=build_parameter, strict_enums=strict_enums)
     return builder.build(schema)
 
 
-def iter_schemas(schema: Schema) -> Iterable[Tuple[str, Any]]:
+def iter_schemas(schema: Schema, strict_enums: Optional[bool] = True) -> Iterable[Tuple[str, Any]]:
     """
     Build zero or more JSON schemas for a marshmallow schema.
 
     Generates: name, schema pairs.
 
     """
-    builder = Schemas(build_parameter=build_parameter)
+    builder = Schemas(build_parameter=build_parameter, strict_enums=strict_enums)
     return builder.iter_schemas(schema)
 
 
-def build_parameter(field: Field) -> Mapping[str, Any]:
+def build_parameter(field: Field, **kwargs) -> Mapping[str, Any]:
     """
     Build JSON parameter from a marshmallow field.
 
     """
-    builder = Parameters()
+    builder = Parameters(**kwargs)
     return builder.build(field)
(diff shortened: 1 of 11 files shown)


microcosm_flask/swagger/definitions.py
```diff
@@ -145,12 +153,12 @@ def iter_definitions(definitions, operations):
...
     for error_schema_class in [ErrorSchema, ErrorContextSchema, SubErrorSchema]:
-        yield error_schema_class()
+        yield error_schema_class(), RequestSide.BOTH
 
     # add all request and response schemas
     for operation, obj, rule, func in operations:
-        yield get_request_schema(func)
-        yield get_response_schema(func)
+        yield get_request_schema(func), RequestSide.REQUEST
+        yield get_response_schema(func), RequestSide.RESPONSE
```
afallou, line 161: Step 1: Identify which side on the request the schema is.

microcosm_flask/swagger/definitions.py
```diff
@@ -128,13 +136,13 @@ def add_definitions(definitions, operations):
...
     """
-    for definition_schema in iter_definitions(definitions, operations):
+    for definition_schema, request_side in iter_definitions(definitions, operations):
         if definition_schema is None:
             continue
         if isinstance(definition_schema, str):
             continue
 
-        for name, schema in iter_schemas(definition_schema):
+        for name, schema in iter_schemas(definition_schema, strict_enums=request_side != RequestSide.RESPONSE):
```
afallou, line 145: Step 2: Pass `strict_enums=False` on the response side.

microcosm_flask/swagger/api.py
```diff
@@ -16,30 +17,30 @@
...
     Build JSON schema from a marshmallow schema.
 
     """
-    builder = Schemas(build_parameter=build_parameter)
+    builder = Schemas(build_parameter=build_parameter, strict_enums=strict_enums)
     return builder.build(schema)
 
 
-def iter_schemas(schema: Schema) -> Iterable[Tuple[str, Any]]:
+def iter_schemas(schema: Schema, strict_enums: Optional[bool] = True) -> Iterable[Tuple[str, Any]]:
```
afallou, line 29: Step 3: Pass `strict_enums` to the OpenAPI schema builder.

microcosm_flask/swagger/parameters/enum.py
```diff
@@ -38,9 +58,15 @@ def parse_type(self, field: Field) -> str:
...
+    def _parse_enum_values(self, field: Field) -> Sequence:
         enum = getattr(field, "enum", None)
         return [
             choice.value if field.by_value else choice.name  # type: ignore
             for choice in enum
         ]
+
+    def parse_enum_values(self, field: Field) -> Optional[Sequence]:
+        if not self.strict_enums:
+            return None
```
afallou, line 70: Step 4: Output for an enum is an enum only with `strict_enums=True`.

rh-glob: nit: Think code would be simpler if it matched your comment
```
if self.strict_enums:
     return self._parse_enum_value(field)
return None
```

afallou:
@tonybajan-globality I believe I could still leave the `enum` key in the definition, and Bravado wouldn't validate it as a enum so long as the `"format": "enum"` piece isn't there. We discussed that possibility, leaving the enum values in for documentation purposes, and ultimately decided against it. It would only add value in the context of a large company where the shared enums aren't readily accessible some other way.

PR #280, opened 2020-06-04
Enable optional health checks

GabeLambda:
Adds the ability to cordon off essential health checks from other
"diagnostic" type of checks.

The health endpoint is intended to be quick, non-intrusive, and accurate
with regards to providing a minimal guarantee of functionality for a
service. However, because there was only one avenue for adding any type of
"status" check here, the endpoint has trended to be slower over time.

This will allow those slower checks to be seen on-demand, but not
interfere with the default checks used by monitoring systems.

This also doesn't change any previous checks; services will have to
"opt-in" to this.

The main impetus here was noticing that some services have extensive
migration histories, causing some checks to hit nearly 100ms(!).

### DIFF
--- a/microcosm_flask/conventions/health.py
+++ b/microcosm_flask/conventions/health.py
@@ -6,13 +6,16 @@ using HTTP 200/503 status codes to indicate healthiness.
 
 """
 from distutils.util import strtobool
+from itertools import chain
+from typing import Any, Dict
 
+from marshmallow import Schema, fields
 from microcosm.api import defaults
 
 from microcosm_flask.audit import skip_logging
 from microcosm_flask.conventions.base import Convention
 from microcosm_flask.conventions.build_info import BuildInfo
-from microcosm_flask.conventions.encoding import make_response
+from microcosm_flask.conventions.encoding import load_query_string_data, make_response
 from microcosm_flask.errors import extract_error_message
 from microcosm_flask.namespaces import Namespace
 from microcosm_flask.operations import Operation
@@ -39,7 +42,7 @@ class HealthResult:
         }
 
     @classmethod
-    def evaluate(cls, func, graph):
+    def evaluate(cls, func, graph) -> "HealthResult":
         try:
             result = func(graph)
             return cls(result=result)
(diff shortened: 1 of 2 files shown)


microcosm_flask/conventions/health.py
```diff
@@ -60,49 +63,64 @@ class Health:
     def __init__(self, graph, include_build_info=True):
         self.graph = graph
         self.name = graph.metadata.name
+        self.checks = dict()
+
         if include_build_info:
-            self.checks = dict(
+            self.optional_checks = dict(
```
GabeLambda, line 69: A bit more of a reach, but I figured if were going to say "/health is *only* essential checks", then these wouldn't make the cut.

microcosm_flask/conventions/health.py
```diff
@@ -60,49 +63,64 @@ class Health:
...
-                for key in sorted(checks.keys())
+                key: check_results[key].to_dict()
+                for key in sorted(check_results.keys())
             }
+
         return dct
 
 
+class RetrieveHealthSchema(Schema):
+    full = fields.Boolean()
```
rh-glob, line 108: What does this default to? None?

microcosm_flask/tests/conventions/test_health.py
```diff
@@ -123,3 +123,76 @@ def fail(graph):
             },
         },
     })))
+
+
+def test_health_check_optional_custom_check():
```
rh-glob, line 128: These seem like they could be parameterized. Also a test explicitly setting full to False would be a nice to have

microcosm_flask/conventions/health.py
GabeLambda: Yeah.

microcosm_flask/tests/conventions/test_health.py
GabeLambda: Sure.

microcosm_flask/conventions/health.py
```diff
@@ -60,49 +63,63 @@ class Health:
     def __init__(self, graph, include_build_info=True):
         self.graph = graph
         self.name = graph.metadata.name
+        self.optional_checks = dict()
+        self.checks = dict()
+
         if include_build_info:
-            self.checks = dict(
+            self.checks.update(dict(
```
GabeLambda, line 70: I decided to keep these as default checks for the moment to avoid possible disruption. These are used pretty frequently, and they're relatively cheap to begin with.

We'll always be able to remove later.

PR #283, opened 2020-06-17
Fix recursive schemas

pokey:
When trying to generate swagger spec for recursive schema definitions, we'd previously get a max recursion depth error because of an infinite descent.

### DIFF
--- a/microcosm_flask/swagger/schemas.py
+++ b/microcosm_flask/swagger/schemas.py
@@ -104,11 +104,9 @@ class Schemas:
 
         for name, field in self.iter_fields(schema):
             if isinstance(field, Nested):
-                yield self.to_tuple(field.schema)
                 yield from self.iter_schemas(field.schema)
 
             if isinstance(field, List) and isinstance(field.inner, Nested):
-                yield self.to_tuple(field.inner.schema)
                 yield from self.iter_schemas(field.inner.schema)
 
     def to_tuple(self, schema: Schema) -> Tuple[str, Any]:
--- a/microcosm_flask/swagger/schemas.py
+++ b/microcosm_flask/swagger/schemas.py
@@ -6,7 +6,7 @@ from typing import (
     Any,
     Callable,
     Iterable,
-    List as TypingList,
+    List as ListType,
     Mapping,
     Tuple,
     Type,
@@ -39,7 +39,7 @@ class Schemas:
 
         # XXX This will break if this class is ever instantiated and then has
         # `build` called more than once
-        self.seen_schemas: TypingList[Type[Schema]] = []
+        self.seen_schemas: ListType[Type[Schema]] = []
 
     def build(self, schema: Schema) -> Mapping[str, Any]:
         """
--- a/microcosm_flask/swagger/schemas.py
+++ b/microcosm_flask/swagger/schemas.py
@@ -2,11 +2,11 @@
 Generate JSON Schema for Marshmallow schemas.
 
 """
-import typing.List
 from typing import (
     Any,
     Callable,
     Iterable,
+    List as TypingList,
     Mapping,
     Tuple,
     Type,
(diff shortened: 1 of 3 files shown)


microcosm_flask/swagger/schemas.py
```diff
@@ -86,6 +92,11 @@ def iter_schemas(self, schema: Schema) -> Iterable[Tuple[str, Any]]:
         if not schema:
             return
 
+        if type(schema) in self.seen_schemas:
```
pokey, line 95: I think this is ok.  We'll never have a schema instantiated twice and mean different things, right?  cc/ @dumpstate @afallou @GabeLambda

microcosm_flask/swagger/schemas.py
```diff
@@ -6,8 +6,10 @@
     Any,
     Callable,
     Iterable,
+    List as TypingList,
```
pokey, line 9: Best practices here?  Conflicts with `marshmallow.fields.List` cc/ @marcglobality @GabeLambda

microcosm_flask/swagger/schemas.py
GabeLambda: If you did, that would seem like a PEBCAK issue.

microcosm_flask/swagger/schemas.py
GabeLambda: `ListType`, maybe.

microcosm_flask/tests/swagger/test_api.py
```diff
@@ -64,3 +64,18 @@ def test_schema_generation_non_strict_enums():
...
         ],
     })))
+
+
+def test_schema_iteration_recursive():
+    # BEWARE: Using `next(...)` instead of `list(...)[0]` here wouldn't
+    # actually test what we want, because it would just grab the first item and
+    # not trigger the infinite recursion.
+    schemas = list(iter_schemas(RecursiveSchema()))
+    name, schema = schemas[0]
```
rh-glob, line 74: nit: remove unused var or add an assert on the name

microcosm_flask/swagger/schemas.py
```diff
@@ -35,6 +37,10 @@ def __init__(
         self.build_parameter = build_parameter
         self.strict_enums = strict_enums
 
+        # XXX This will break if this class is ever instantiated and then has
```
afallou, line 40: I think you can remove the `XXX`

microcosm_flask/swagger/schemas.py
marcglobality: in `microcosm_flask/tests/conventions/fixtures.py` they import like `from marshmallow import fields` and do `fields.List`. If you were doing the same here there would be no conflict. In any case I think it should be uniformised at least for the same repo, whichever pattern is decided

In general, I'm more inclined to have `{name_of_library}_{module}`. I had similar imports in the past as in

```
from sklearn.metrics import precision as sklearn_precision
from my_module import precision
```
so I'd prefer your original choice which is the analogous when using PascalCase

microcosm_flask/swagger/schemas.py
pokey: @afallou Why remove the `XXX`?  Because we have no intention of ever fixing?  Makes sense I guess.  Maybe `NB`?

microcosm_flask/swagger/schemas.py
```diff
@@ -86,18 +92,21 @@ def iter_schemas(self, schema: Schema) -> Iterable[Tuple[str, Any]]:
...
         for associated_schema in getattr(schema, associated_schemas_attr_name(schema.__class__), {}).values():
             yield self.to_tuple(associated_schema())
 
         for name, field in self.iter_fields(schema):
             if isinstance(field, Nested):
-                yield self.to_tuple(field.schema)
                 yield from self.iter_schemas(field.schema)
 
             if isinstance(field, List) and isinstance(field.inner, Nested):
-                yield self.to_tuple(field.inner.schema)
```
pokey, line 100: Somewhat surprisingly, this line and the one above aren't necessary, and in fact were resulting in every nested schema being output twice!  Was being masked by the fact that the schemas are just put [directly into a dict](https://github.com/globality-corp/microcosm-flask/blob/a36f8372caa87b1b40e92b0a297789c7c0289c84/microcosm_flask/swagger/definitions.py#L146).  Caught this issue when adding one final sanity check unit test cc/ @GabeLambda @rh-glob @afallou

GabeLambda: Oops.

PR #287, opened 2020-09-15
Remove jaeger/tracing references

GabeLambda:
These spans/traces weren't used, and will be removed from microcosm as a
whole.

Note: This used to rely on an as-yet released version of microcosm: https://github.com/globality-corp/microcosm/pull/78
However, it's probably an easier transition if we merely remove dependencies on it first, then upgrade to the version that naturally lacks these.


microcosm_flask/tests/test_metrics.py
```diff
@@ -120,7 +118,6 @@ def foo():
             tags=[
                 "endpoint:foo.search.v1",
                 "backend_type:microcosm_flask",
-                "classifier:404",
```
GabeLambda, line 123: Unrelated, but the tests were being skipped by default, and wouldn't pass when enabled.

dumpstate (approved):

afallou (approved):

PR #291, opened 2020-10-02
normalizing status codes in route metrics - reduce cardinality

BarakBD-Globality:
https://globality.atlassian.net/browse/DEVOPS-2539

### DIFF
--- a/microcosm_flask/metrics_classifier.py
+++ b/microcosm_flask/metrics_classifier.py
@@ -14,8 +14,15 @@ class StatusCodeClassifier(Classifier):
     """
     def label_result(self, result):
         _, status_code, _ = parse_response(result)
-        return str(status_code)
+        return self.normalize_status_code(status_code)
 
     def label_error(self, error):
         status_code = extract_status_code(error)
-        return str(status_code)
+        return self.normalize_status_code(status_code)
+
+    # Return only the first digit of the status code followed by xx as a string
+    # Example: 403 becomes "4xx"
+    # This will reduce the cardinality of metrics collected which will reduce cost
+    # For detailed status code errors we have logs
+    def normalize_status_code(self, status_code: int) -> str:
+        return str(status_code)[0] + "xx"
--- a/setup.py
+++ b/setup.py
@@ -40,11 +40,18 @@ setup(
         "sentry": "sentry-sdk>=0.14.4",
         "spooky": "spooky>=2.0.0",
         "test": [
+            "nose>=1.3.7",
             "sentry-sdk>=0.14.4",
             "PyHamcrest",
             "coverage",
-            "parameterized"
+            "parameterized",
         ],
+        "lint": [
+            "flake8",
+            "flake8-print",
+            "flake8-logging-format",
+            "flake8-isort"
+        ]
     },
     setup_requires=[
         "nose>=1.3.7",
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,7 +14,7 @@
 defaults: &defaults
   working_directory: ~/repo
   docker:
-    - image: ${AWS_ECR_DOMAIN}/globality-build:2020.36.100
+    - image: ${AWS_ECR_DOMAIN}/globality-build:2020.39.10
       aws_auth:
         aws_access_key_id: ${AWS_ACCESS_KEY_ID}
         aws_secret_access_key: ${AWS_SECRET_ACCESS_KEY}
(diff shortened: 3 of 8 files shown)


setup.py
```diff
@@ -40,11 +40,20 @@
         "sentry": "sentry-sdk>=0.14.4",
         "spooky": "spooky>=2.0.0",
         "test": [
+            "nose>=1.3.7",
             "sentry-sdk>=0.14.4",
             "PyHamcrest",
             "coverage",
-            "parameterized"
+            "parameterized",
+            "microcosm-metrics>=2.5.1"
```
BarakBD-Globality, line 48: See test setup in microcosm_flask/tests/test_metrics.py
If we do not import microcosm-metrics - metrics tests are skipped.

microcosm_flask/metrics_classifier.py
```diff
@@ -14,8 +14,17 @@ class StatusCodeClassifier(Classifier):
...
         status_code = extract_status_code(error)
-        return str(status_code)
+        return self.normalize_status_code(status_code)
+
+    # Return only the first digit of the status code followed by xx as a string
+    # Example: 403 becomes "4xx"
+    # This will reduce the cardinality of metrics collected which will reduce cost
+    # For detailed status code errors we have logs
+    def normalize_status_code(self, status_code: int) -> str:
+        if isinstance(status_code, int):
```
GabeLambda, line 28: This type-hint is redundant. If you *are* getting multiple types, that should be reflected in the signature above.

setup.py
GabeLambda: There were no changes to the tests?

microcosm_flask/metrics_classifier.py
```diff
@@ -14,8 +14,17 @@ class StatusCodeClassifier(Classifier):
...
+            return str(status_code)[:1] + "xx"
```
GabeLambda, line 29: If you just need the first character, why not `[0]`?

setup.py
BarakBD-Globality: Not sure I undrstand the question. 
I can pip install `microcosm-metrics` in the `test` section in `entrypoint.sh` if that makes more sense

BarakBD-Globality: There are changes to the tests

microcosm_flask/metrics_classifier.py
BarakBD-Globality: Fair enough

microcosm_flask/metrics_classifier.py
BarakBD-Globality: just being over-safe, as this is a library.
But I agree, it can be removed.

setup.py
GabeLambda: Ah, never mind.

microcosm_flask/metrics_classifier.py
GabeLambda: I mean, argument being that this method is technically safer, as you won't IndexError out. But I imagine you could also make that check earlier to see if the input results in an empty string or not.

microcosm_flask/metrics_classifier.py
GabeLambda: It is a library, but how it's getting the status code is from `parse_response`, another internal function.

We really should have more typehints in this repo, but the expected values of the responses should mean you'd always get an int here.

GabeLambda:
I'm assuming the main impetus for this change is cost, but we...somehow don't care about the exact response codes?

This might break the existing dashboards, but I assume you're going to fix that as well.

GabeLambda (approved):

amenut:
> I'm assuming the main impetus for this change is cost, but we...somehow don't care about the exact response codes?
> 
> This might break the existing dashboards, but I assume you're going to fix that as well.

Coming in late to this, but I'm not finding dashboards where we're breaking down response codes on a per that are actually used? The dashboards that have been of interest have charted aggregate sums of entire classes (2xx, 4xx, etc).

Even so, adding back the exact response codes isn't going to add a lot back in terms of metrics load, when compared to `NaN` values output by the statsd exporter.

Randomly selecting an EC2 instance from service-dev:
```
arnaud.menut@ip-10-40-82-159:~$ curl -s localhost:9102/metrics | wc -l
1742
arnaud.menut@ip-10-40-82-159:~$ curl -s localhost:9102/metrics | grep route_call_count | wc -l
120
arnaud.menut@ip-10-40-82-159:~$ curl -s localhost:9102/metrics | grep NaN | wc -l
813
```

Route call counts are 6% of the metrics, and are all valid values.
`NaN` values are 46.67% out of the output.

tl;dr: adding back exact response codes is likely negligible as a cost consideration.

PR #302, opened 2021-01-13
[Devops-2968] replace human creds with api token

sonuGlobality:
Replaces user/pass with pypi token that will prevent build failures on passwords rotation

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -28,7 +28,8 @@ elif [ "$1" = "lint" ]; then
    # Install standard linting dependencies; YMMV
    pip --quiet install \
        .[lint] flake8 flake8-print flake8-logging-format flake8-isort
-   exec flake8 ${NAME}
+   flake8 ${NAME}
+    exit
 elif [ "$1" = "typehinting" ]; then
    # Install standard type-linting dependencies
    pip --quiet install mypy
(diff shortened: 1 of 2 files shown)


entrypoint.sh
```diff
@@ -28,7 +28,8 @@ elif [ "$1" = "lint" ]; then
    # Install standard linting dependencies; YMMV
    pip --quiet install \
        .[lint] flake8 flake8-print flake8-logging-format flake8-isort
-   exec flake8 ${NAME}
+   flake8 ${NAME}
+    exit
```
simpleigh-globality, line 32: What does this do? Also seems to be misaligned?

simpleigh-globality: Not sure it's needed TBH as we're in an `if` statement and the script terminates afterwards.

simpleigh-globality:
```suggestion
```

tonybajan-globality (approved):
lgtm once leigh's note is addressed

entrypoint.sh
sonuGlobality: I don't see any need for exit here but la me ask @marcglobality 

@marcglobality 
This change was introduced by you in this PR : https://github.com/globality-corp/globality-build/pull/480/files
Any comment?

simpleigh-globality: Created https://github.com/globality-corp/globality-build/pull/483

sonuGlobality: merged

PR #305, opened 2021-01-21
Add byte order mark when outputting CSV for Excel

tonybajan-globality:
By default, Excel will open text CSV files as a legacy encoding, but it
will detect a BOM as an in-band indiction that the file is encoded as
UTF-8.

When formatting CSVs, use Python's utf-8-sig codec to add the BOM
to the response body.

### DIFF
--- a/microcosm_flask/formatting/csv_formatter.py
+++ b/microcosm_flask/formatting/csv_formatter.py
@@ -9,6 +9,7 @@ from werkzeug import Response
 from werkzeug.utils import get_content_type
 
 from microcosm_flask.formatting.base import BaseFormatter
+from microcosm_flask.formatting.encoding import UTF_8, UTF_8_SIG
 
 
 class CSVFormatter(BaseFormatter):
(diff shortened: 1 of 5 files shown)


microcosm_flask/formatting/csv_formatter.py
```diff
@@ -23,6 +26,17 @@ def build_headers(self, headers, **kwargs):
...
 
+    def build_response(self, response_data):
+        response = Response(
+            self.format(response_data),
+            content_type=get_content_type(self.content_type, 'utf-8')
+        )
+
+        # start the output with U+FEFF BYTE ORDER MARK
+        # to signal to Excel to import the text file as UTF-8 rather than a legacy encoding
+        response.charset = "utf-8-sig"
```
rh-glob, line 37: Should this move to a constant var? or is this value available from any of `microcosm_flask` dependencies 🤔

rh-glob (approved):

microcosm_flask/formatting/csv_formatter.py
tonybajan-globality: It's unfortunately not a constant in the stdlib `codecs` or `encodings.utf_8_sig` - will add some constants here

isabelcosta-globality (approved):
🤓

PR #310, opened 2021-08-23
Release/3.6.4

mgoldman-globality:
adding AsciiEncodedStringField


setup.py
```diff
@@ -21,6 +21,7 @@
     python_requires=">=3.6",
     keywords="microcosm",
     install_requires=[
+        "Flask<2",
```
sammy-globality, line 24: What breaks with Flask>=2?

adeleinr: I have the same question

adeleinr (approved):

setup.py
adeleinr: Sorry that question was missing context, what breaks if you dont use Frask <2 is the tests for the PR Tony just merged before this one. I talked to him and he checked his tests and thinks that flask >=2 is whats breaking them so he suggested to downgrade flast. But, I am not sure exactly how it is breaking those tests, if it is worth noting it at least in the PR.

mgoldman-globality: I'll look into it and see if I can figure out why it's breaking them

adeleinr: Thanks, lets not block this PR on that, but if you do find out let us know

sammy-globality (approved):

PR #313, opened 2021-08-31
Update landing page to use optional full health check param

sethisernhagen:
- update landing page to show optional full health check data

Why?

The default health endpoint returns a minimum set of health data. This
is generally used to make sure a service is "up".

The landing page is meant to be a human consumable resource showing
all health and config data. Since this view is meant for humans to see
a complete view of a service it makes sense to show full health check
data.

### DIFF
--- a/microcosm_flask/conventions/landing.py
+++ b/microcosm_flask/conventions/landing.py
@@ -85,7 +85,7 @@ def configure_landing(graph):   # noqa: C901
         """
         config = graph.config_convention.to_dict()
         env = get_env_file_commands(config, graph.metadata.name)
-        health = graph.health_convention.to_dict()
+        health = graph.health_convention.to_dict(full=True)
         properties = get_properties_and_version()
         swagger_versions = get_swagger_versions()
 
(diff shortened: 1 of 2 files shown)


JP-Globality (approved):

microcosm_flask/conventions/landing.py
```diff
@@ -85,7 +85,7 @@ def render_landing_page():
         """
         config = graph.config_convention.to_dict()
         env = get_env_file_commands(config, graph.metadata.name)
-        health = graph.health_convention.to_dict()
+        health = graph.health_convention.to_dict(full=True)
```
rh-glob, line 88: Is this needed? You are hardcoding `full=True` in the template so doesn't look like this addition will be used during the render?

tonybajan-globality (approved):

microcosm_flask/conventions/landing.py
sethisernhagen: The template change below only updates a link to the health endpoint. I thought it was helpful to update that because it might help people remember the `full` param exists. I doubt anyone knows that this link is even there.

The rendering of the health info happens here server side. This is the real change.

PR #316, opened 2021-09-01
Merge master to develop

sethisernhagen:


sethisernhagen:
This would work but doesn't look right. I think the diff should only include the version bump.

PR #318, opened 2022-01-23
GLOB-61031: Add graph component for catch-all route

afallou:
https://globality.atlassian.net/browse/GLOB-61031

## Why?
Flask's default behavior when receiving a request for a non-existent route is to return a 404.

While that is the obvious right choice for any server returning a response to a web browser, in a backend environment it can help to have the ability to differentiate the cases of a correct path but missing resource (i.e. a `404` on /api/v1/user/<uuid>) from the case of a wrong path.

This is particularly useful in an envionment where service instances are spun up and down often enough that propagation of DNS records can lag behind – meaning that a load balancer can end up sending a request intended for one service to another. A different HTTP code for a wrong path allows tuning the load balancer behavior to mark the corresponding DNS record as stale when receiving that HTTP code.

**Note:** I'm aware that this is a slight abuse of HTTP codes; the practical benefits make this worth it.

## What?
Add `catchall_convention` component adding a catch-all Flask route.
Flask/Werkzeug's routing priority order is documented [here](https://github.com/pallets/werkzeug/blob/347291802fcf89bc89660cc9dc62eb1303337bc2/src/werkzeug/routing.py#L1133). This catch-all route will always have fewer static parts than any other, and as a result should come last in priority. This means it won't override any existing route.

## Testing
* [x] Deployed on a dev environment and ran automation tests, to verify this has no adverse effects.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -183,9 +183,24 @@ jobs:
       - attach_workspace:
           at: ~/repo
       - run:
-          name: Deploy
+          name: Override version
           command: |
-            echo "Not publishing package!"
+            sed -i -e "s/^\(version = \".*\)\"$/\1.dev${CIRCLE_BUILD_NUM}\"/" setup.py
+      - run:
+          name: Authenticate
+          command: |
+            echo "[distutils]" > ~/.pypirc
+            echo "index-servers =" >> ~/.pypirc
+            echo "    jfrog" >> ~/.pypirc
+            echo >> ~/.pypirc
+            echo "[jfrog]" >> ~/.pypirc
+            echo "repository:https://globality.jfrog.io/globality/api/pypi/pypi" >> ~/.pypirc
+            echo "username:$JFROG_USERNAME" >> ~/.pypirc
+            echo "password:$JFROG_PASSWORD" >> ~/.pypirc
+            echo >> ~/.pypirc
+      - run:
+          name: Publish
+          command: python setup.py sdist upload -r jfrog
 
   publish_library:
     <<: *defaults
@@ -255,6 +270,9 @@ workflows:
             - test
             - lint
             - typehinting
+          filters:
+            branches:
+              ignore: master
       - typehinting:
           context: 
           - Globality-Common
--- a/microcosm_flask/conventions/catchall.py
+++ b/microcosm_flask/conventions/catchall.py
@@ -1,0 +1,26 @@
+"""
+Flask's default behavior is to return a 404 for a route the app can't match.
+For backend services that are not queried directly by a front-end it can be useful
+to specify a different HTTP code, to differentiate between a nonexistent resource
+on a well-understood path (e.g. GET /user/123) and a nonexistent (e.g. GET /ursr).
+
+In particular, this can allow a reverse-proxy to identify cases where it is routing a request
+based on an outdated DNS entry.
+
+
+"""
+from microcosm.api import binding, defaults, typed
+
+
+@binding("catchall_convention")
+@defaults(
+    fallback_http_code=typed(int, default_value=421),
+)
+def configure_catchall_convention(graph):
+    fallback_http_code = graph.config.catchall_convention.fallback_http_code
+
+    @graph.flask.route("/<path:path>", methods=["GET", "POST", "PATCH", "PUT", "DELETE"])
+    def catch_all(path):
+        return "Unknown path", fallback_http_code
+
+    return catch_all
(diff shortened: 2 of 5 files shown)


microcosm_flask/swagger/parameters/enum.py
```diff
@@ -59,7 +59,7 @@ def parse_type(self, field: Field) -> str:
             raise Exception(f"Cannot infer enum type for field: {field.name}")
 
     def _parse_enum_values(self, field: Field) -> Sequence:
-        enum = getattr(field, "enum", None)
+        enum = getattr(field, "enum")
```
afallou, line 62: Unrelated change to fix typehinting. Not that in the `None` case line 63 would have errored anyway, so this change doesn't introduce a new risk – this code is only reached when `self.strict_enums` is true (see line 69), in which case the `"enum"` attribute should be present.

dumpstate (approved):

microcosm_flask/swagger/parameters/enum.py
dmytrostarosud-globality: `field.enum` then?

microcosm_flask/conventions/catchall.py
```diff
@@ -0,0 +1,26 @@
...
+@binding("catchall_convention")
+@defaults(
+    fallback_http_code=typed(int, default_value=421),
+)
+def configure_catchall_convention(graph):
+    fallback_http_code = graph.config.catchall_convention.fallback_http_code
+
+    @graph.flask.route("/<path:path>", methods=["GET", "POST", "PATCH", "PUT", "DELETE"])
+    def catch_all(path):
+        return "Unknown path", fallback_http_code
```
simpleigh-globality, line 24: Will this be logged in the same way as other requests?

simpleigh-globality (approved):

microcosm_flask/conventions/catchall.py
afallou: After double-checking: I believe so – the audit logging setup is for all Flask routes.

afallou: Now it will be – I added the audit decorator on this route.

microcosm_flask/swagger/parameters/enum.py
```diff
@@ -59,10 +59,9 @@ def parse_type(self, field: Field) -> str:
             raise Exception(f"Cannot infer enum type for field: {field.name}")
 
     def _parse_enum_values(self, field: Field) -> Sequence:
-        enum = getattr(field, "enum", None)
         return [
             choice.value if field.by_value else choice.name  # type: ignore
-            for choice in enum
+            for choice in field.enum
```
afallou, line 64: Unrelated change to fix typehinting. Not that in the `None` case line 63 would have errored anyway, so this change doesn't introduce a new risk. This code is only reached when self.strict_enums is true (see line 69), in which case the "enum" attribute should be present.

microcosm_flask/conventions/catchall.py
mlcohen: @afallou few things:

1. can the audit include the unhandled route?
2. not sure what `success` means here. Seems like it should be `false` in this particular case
3. tiny nit: can we adjust the format of `elapased_time`? Bring it more up to date like other elapsed time field where the default unit of measurement is `ms` instead of `s`.

mlcohen (approved):
added minor comments

microcosm_flask/conventions/catchall.py
```diff
@@ -0,0 +1,27 @@
...
+based on an outdated DNS entry, and to directly retry the request on a different instance.
+
+
+"""
+from microcosm.api import binding, defaults, typed
+
+
+@binding("catchall_convention")
+@defaults(
+    fallback_http_code=typed(int, default_value=421),
```
mlcohen, line 17: based on comments [here](https://github.com/globality-corp/serviceproxy/pull/170#discussion_r790676664), the default should be updated to be `501`

microcosm_flask/conventions/catchall.py
afallou: @mlcohen I changed the logic to raise an `HTTPException`, that way the audit log will return `success: false` and include the response body (I added the path in it).
As for 3. the `elapsed_time` is already in ms, it's just fast 😄 .

mlcohen (approved):

.circleci/config.yml
```diff
@@ -183,9 +183,24 @@ jobs:
...
+            echo "    jfrog" >> ~/.pypirc
+            echo >> ~/.pypirc
+            echo "[jfrog]" >> ~/.pypirc
+            echo "repository:https://globality.jfrog.io/globality/api/pypi/pypi" >> ~/.pypirc
+            echo "username:$JFROG_USERNAME" >> ~/.pypirc
+            echo "password:$JFROG_PASSWORD" >> ~/.pypirc
+            echo >> ~/.pypirc
+      - run:
+          name: Publish
+          command: python setup.py sdist upload -r jfrog
```
afallou, line 203: This comes from merging master into this branch (see https://github.com/globality-corp/microcosm-flask/pull/319)

PR #320, opened 2022-01-26
Release version 3.7.0

afallou:
Releases https://github.com/globality-corp/microcosm-flask/pull/318

Note: the changes from https://github.com/globality-corp/microcosm-flask/pull/317 are included here because the git history is a little messy, but they've already been release into `3.6.5`.


afallou:
Released via another PR

PR #322, opened 2022-03-15
Release version 4.0.0

mgoldman-globality:
https://globality.atlassian.net/browse/GLOB-61300

removes flask<2 constraint https://github.com/globality-corp/microcosm-flask/pull/321

##### Note:
This is not backwards-compatible with flask<2. To use this in services with a flask or werkzeug  dependency both libraries must be updated to >=2.0


.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 3.6.5
+current_version = 4.0.0
```
marcglobality, line 2: as commented on slack, we should probably do 3.7.0

marcglobality (approved):

.bumpversion.cfg
ronanm-globality: I suspect it might run into issues bumping Flask as a major version, without also bumping `microcosm-flask` as a major version.

Mostly concerned upgrades from `3.X` will probably run into a conflict with the `Flask<2` constraint in other `microcosm-*` and `marquez` libraries

marcglobality: so what's next action from here?

sethisernhagen (approved):
LGTM. Is it good time to also delete the dev branch and not do git-flow anymore? Will be still be able to release 3.x fixes?

PR #331, opened 2022-08-01
Optional non-strict OpenAPI definition for enums

dumpstate:
For some enums, we'd like to keep OpenAPI protocol open and defer to service level validation.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 3.7.1
+current_version = 3.8.0
 commit = False
 tag = False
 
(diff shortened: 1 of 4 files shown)


ronanm-globality:
Was thinking about how this interacts with the `EnumField` marshmallow field we define, but I think it "just works" right? As it relies on a `return self.enum(value)`, which in our case is dynamically returning a value for anything in the right format.

ronanm-globality (approved):

dumpstate:
> Was thinking about how this interacts with the `EnumField` marshmallow field we define, but I think it "just works" right? As it relies on a `return self.enum(value)`, which in our case is dynamically returning a value for anything in the right format.

Yes, that's correct.

JP-Globality (approved):

afallou (approved):

tonybajan-globality (approved):

PR #332, opened 2022-09-08
update add responses to use getdoc

swahl-globality:
Descriptions of paths and responses that are pulled from the docstring are behaving inconsistently.

In the path "/deferral_feedback/{deferral_feedback_id}" for picasso, patch has the following description
```
"patch": {
        "description": "Support re-encryption by enforcing that every update triggers a\nnew encryption call, even if the the original call does not update\nthe encrypted field.",
```
whereas the response description below is cut off at the new line:
```
responses": {
          "200": {
            "description": "Support re-encryption by enforcing that every update triggers a",
```

The two sections in `microcosm-flask/swagger/definitions.py` use different methods to get the docstring. Converting to use `getdoc` ensures the same behavior between the two sections.

### DIFF
--- a/microcosm_flask/errors.py
+++ b/microcosm_flask/errors.py
@@ -131,11 +131,8 @@ def make_json_error(error):
     # Flask will not log user exception (fortunately), but will log an error
     # for exceptions that escape out of the application entirely (e.g. if the
     # error handler raises an error)
-    error_logger.debug("Handling {} error: {}".format(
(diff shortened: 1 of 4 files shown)


ronanm-globality (approved):

shilpakatta-globality (approved):

sstrasser-globality (approved):
🎂

PR #337, opened 2022-10-05
Remove werkzeug constraint

tomaszpietruszka-globality:
Updating `werkzeug` to 2.2 - which involves a change to its router mechanism - causes 2 tests to fail:

1. `test_alias.py` - Aliases don't seem to work, unless configured before `configure_crud`? Changed the test setup, but perhaps changing the actual behavior would be the right thing to do?
2. `test_landing.py` - the "landing convention" appears to be broken?


marcglobality:
> Updating `werkzeug` to 2.2 - which involves a change to its router mechanism - causes 2 tests to fail:
> 
> 1. `test_alias.py` - Aliases don't seem to work, unless configured before `configure_crud`? Changed the test setup, but perhaps changing the actual behavior would be the right thing to do?
> 2. `test_landing.py` - the "landing convention" appears to be broken?

Re2: the problem comes from [here](https://github.com/pallets/flask/pull/4577/files) where they moved from warning to error. Long story short, `before_request` doesn't seem to be working for us once the application is running...

microcosm_flask/tests/conventions/test_alias.py
```diff
@@ -59,8 +46,8 @@ def setup(self):
         self.ns = Namespace(
             subject=Person,
         )
-        configure_crud(self.graph, self.ns, PERSON_MAPPINGS)
         configure_alias(self.graph, self.ns, PERSON_MAPPINGS)
+        configure_crud(self.graph, self.ns, PERSON_MAPPINGS)
```
tomaszpietruszka-globality, line 50: As noted by @ronanm-globality, docs elsewhere tell us we _must_ configure these in the "old" order, so the new changes seem a bit wrong - or at least we should update the docs.

(also unfortunate: these things seem to just quietly not work when done in the wrong order - which could make the migration frustrating for users

marcglobality (approved):

ronanm-globality (approved):

PR #338, opened 2022-10-05
GLOB-68575: Add schema, schema fields, and query params descriptions for improved API doc generation

GaryGlobality:
This change enhances the automatic swagger API documentation mechanism by:

- Adding an optional `description` property to the `EndpointDefinition` class such that we can include end-point descriptions in the documentation if provided.
- Including marshmallow Field `metadata.description` in the documentation if provided for query parameters, request and response schema fields.
- Adding such descriptions to the pagination fields like `offset`, `limit`, and `count`. Also documenting the default header `X-Response-Skip-Null`
- Including marshmallow Schema `__doc__` strings in the documentation if provided.

The `black` formatter was also automatically applied.

### DIFF
--- a/microcosm_flask/paging.py
+++ b/microcosm_flask/paging.py
@@ -55,8 +55,12 @@ def identity(x):
 # NB: lots of code currently uses `PageSchema` to refer to `OffsetLimitPageSchema`
 # keeping this (mis)naming for backwards compatibilty
 class PageSchema(Schema):
-    offset = fields.Integer(missing=None)
-    limit = fields.Integer(missing=None)
+    offset = fields.Integer(
+        missing=None, metadata={"description": "The pagination starting offset."}
+    )
+    limit = fields.Integer(
+        missing=None, metadata={"description": "The pagination limit."}
+    )
 
 
 class OffsetLimitPageSchema(PageSchema):
@@ -70,6 +74,7 @@ class PaginatedList:
     Includes HAL-style links (e.g for the current or next page)
 
     """
+
     def __init__(self, items, _page, _ns, _operation, _context):
         self.items = items
         self._page = _page
@@ -89,10 +94,7 @@ class PaginatedList:
         """
         links = Links()
         links["self"] = Link.for_(
-            self._operation,
-            self._ns,
-            qs=self._page.to_items(),
-            **self._context
+            self._operation, self._ns, qs=self._page.to_items(), **self._context
         )
         return links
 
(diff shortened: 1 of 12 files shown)


microcosm_flask/paging.py
```diff
@@ -316,11 +314,25 @@ def make_paginated_list_schema_class(cls, ns, item_schema):
...
-            limit = fields.Integer(required=True)
-            count = fields.Integer(required=True)
-            items = fields.List(fields.Nested(item_schema), required=True)
-            _links = fields.Raw()
+            offset = fields.Integer(
+                required=True,
+                metadata={"description": "The pagination starting offset."},
+            )
+            limit = fields.Integer(
+                required=True, metadata={"description": "The pagination limit."}
```
jpringle-globality, line 322: Nit: comma after second arg
```suggestion
                required=True,
                metadata={"description": "The pagination limit."},
```

microcosm_flask/swagger/parameters/base.py
```diff
@@ -73,6 +76,10 @@ def parse_description(self, field: Field) -> Optional[str]:
         Parse the description for the field, if any.
 
         """
+        if "metadata" in field.metadata:
+            metadata = field.metadata.get("metadata")
+            if metadata:
+                return metadata.get("description")
```
jpringle-globality, line 82: Nit: possible simplification

```suggestion
        metadata = field.metadata.get("metadata", {})
        if metadata:
            return metadata.get("description")
```

jpringle-globality: Also, how do we get `field.metadata.metadata.description`? Does this happen?

microcosm_flask/swagger/schemas.py
```diff
@@ -59,14 +61,15 @@ def build(self, schema: Schema) -> Mapping[str, Any]:
...
-            for name, field in fields
-            if field.required and not field.allow_none
+            name for name, field in fields if field.required and not field.allow_none
         ]
 
         if required_fields:
             result["required"] = required_fields
 
+        if schema.__doc__:
+            result["description"] = re.sub(r"\s+", " ", schema.__doc__.strip())
```
jpringle-globality, line 71: this corrects the multiline docstring getting cut off at the `\n` ? Looks good

jpringle-globality (approved):
Looks good to me

microcosm_flask/swagger/parameters/base.py
GaryGlobality: It happens for the fields of marshmallow `Schema`, but I'm not really sure why as we don't nest the metadata ourselves. I handled both cases to be defensive.

microcosm_flask/swagger/schemas.py
GaryGlobality: It's primarily replacing extraneous whitespace in the docstring so that it looks neat in the generated documentation.

mcl-globality (approved):

PR #340, opened 2022-10-18
DEVOPS-3866 Old ticket cleanup context

Blizzard6642:
Changing "- Python-Common" to "- Python-Context" as the previous one …does not exist.

Just updating context for an old ticket


jpratt3000-globality:
no longer needed

PR #348, opened 2023-06-02
WIP Add support for adding x nullable to swagger deft

sstrasser-globality:


sstrasser-globality:
Still need to fix tests and test this a little more.

PR #354, opened 2024-03-14
Add New Marshmallow Field

sstrasser-globality:
Add new field so that we can pass more sophisticated sorting to our APIs. 

Example: `/articles?sort=-created,title` 

Concept pulled from from https://jsonapi.org/format/1.1/#fetching-sorting

### DIFF
--- a/microcosm_flask/fields/order_by_field.py
+++ b/microcosm_flask/fields/order_by_field.py
@@ -35,6 +35,9 @@ class OrderByField(fields.Str):
         - OrderBy(created, OrderDirection.DESC)
         - OrderBy(title, OrderDirection.ASC)
 
+        Note that in this example, Title does not have a +/-. This will give the field a default of sort ASC. This is
+        all in accordance with the JSON API standard.
+
     JSON API standard: https://jsonapi.org/format/1.1/#fetching-sorting
     """
 
--- a/microcosm_flask/fields/order_by_field.py
+++ b/microcosm_flask/fields/order_by_field.py
@@ -1,0 +1,62 @@
+from enum import Enum
+
+from marshmallow import fields
+
+from microcosm_flask.fields import EnumField
+
+
+class OrderDirection(Enum):
+    ASC = "+"
+    DESC = "-"
+
+
+class OrderBy:
+
+    def __init__(self, order_field: Enum, order_dir: OrderDirection):
+        self.order_field = order_field
+        self.order_dir = order_dir
+
+    def __str__(self):
+        # This is used when serializing the object to a string for the URL. It doesn't actually use the _serialize
+        # method below.
+        return self.order_dir.value + self.order_field.name
+
+
+class OrderByField(fields.Str):
+    """
+    OrderByField valued field. Allow for sorting by a set of fields and directions. Most use cases will wrap the field
+    within a QueryStringList field to allow for multiple values, but it is not required. The field requires setting
+    the enum field parameter to allow for users to create their own enum types to use for sorting.
+
+    Example:
+        the url /articles?sort=-created,title
+
+        will create a sort field with two OrderBy entries:
+        - OrderBy(created, OrderDirection.DESC)
+        - OrderBy(title, OrderDirection.ASC)
+
+    JSON API standard: https://jsonapi.org/format/1.1/#fetching-sorting
+    """
+
+    def __init__(self, enum_field, *args, **kwargs):
+        self.enum_field: EnumField = enum_field
+        super().__init__(*args, **kwargs)
+
+    def _serialize(self, value, attr, obj, **kwargs):
+        if value is None:
+            return None
+        # Just return the __str__ from above
+        return str(value)
+
+    def _deserialize(self, value, attr, data, **kwargs):
+        if value.startswith(OrderDirection.DESC.value):
+            direction = OrderDirection.DESC
+            parsed_value = value[1:]
+        elif value.startswith(OrderDirection.ASC.value):
+            direction = OrderDirection.ASC
+            parsed_value = value[1:]
+        else:
+            direction = OrderDirection.ASC
+            parsed_value = value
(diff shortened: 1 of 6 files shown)


microcosm_flask/fields/order_by_field.py
```diff
@@ -0,0 +1,62 @@
...
+        self.order_field = order_field
+        self.order_dir = order_dir
+
+    def __str__(self):
+        # This is used when serializing the object to a string for the URL. It doesn't actually use the _serialize
+        # method below.
+        return self.order_dir.value + self.order_field.name
+
+
+class OrderByField(fields.Str):
```
sstrasser-globality, line 25: One of my original versions had this extend the enum class, but then that gets into an issue when we generate the swagger defts as it only sets the valid enums as those passed in instead of something like "+ENUM". This isn't an issue server side, but breaks client side when using Bravado.

german-glob (approved):
LGTM! One question, how you're going to use `order_field` in SQLAlchemy query?

microcosm_flask/fields/order_by_field.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def _deserialize(self, value, attr, data, **kwargs):
+        if value.startswith(OrderDirection.DESC.value):
+            direction = OrderDirection.DESC
+            parsed_value = value[1:]
+        elif value.startswith(OrderDirection.ASC.value):
+            direction = OrderDirection.ASC
+            parsed_value = value[1:]
+        else:
+            direction = OrderDirection.ASC
```
ronanm-globality, line 59: You used it in the example in the docstring, but it could be good to explicitly call out the `ASC` default as well 👍

ronanm-globality (approved):

microcosm_flask/tests/fields/test_order_by_field.py
```diff
@@ -0,0 +1,141 @@
...
+    )
+
+    assert_that(result["order_by_str"], is_(equal_to(f"+{ForTestEnum.Foo.name}")))
+    assert_that(result["order_by_int"], is_(equal_to(f"-{ForTestIntEnum.Bar.name}")))
+    # assert_that(result["order_by_str_field_list"], is_(equal_to(
+    #     f"+{ForTestEnum.Foo.name},-{ForTestEnum.Bar.name},{ForTestEnum.Bar.name}"
+    # )))
+    # assert_that(result["order_by_int_field_list"], is_(equal_to(
+    #     f"+{ForTestIntEnum.Foo.name},-{ForTestIntEnum.Bar.name},{ForTestIntEnum.Bar.name}"
+    # )))
```
anthony-globality, line 119: can these be enabled?

microcosm_flask/fields/order_by_field.py
sstrasser-globality: Good call.

microcosm_flask/tests/fields/test_order_by_field.py
sstrasser-globality: GAH. Disabled them for testing other test. Will add back in.

sstrasser-globality:
@german-glob The idea would be something like this: 

```        
for thing in better_order_by:
            if thing.order_dir == OrderDirection.DESC:
                query = query.order_by(desc(text(thing.order_field.value)))
            else:
                query = query.order_by(asc(text(thing.order_field.value)))
```

I'm sure there's cleanup I could do on the query and use get_attr or something else fancier,  but you get the general idea. 
https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.order_by

german-glob:
> @german-glob The idea would be something like this:
> 
> ```
> for thing in better_order_by:
>             if thing.order_dir == OrderDirection.DESC:
>                 query = query.order_by(desc(text(thing.order_field.value)))
>             else:
>                 query = query.order_by(asc(text(thing.order_field.value)))
> ```
> 
> I'm sure there's cleanup I could do on the query and use get_attr or something else fancier, but you get the general idea. https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.order_by

Yeah, I'm asking because of the SQL injection. We need to make sure it won't open doors for it. Let's check if order by is sent as param to the DB. Maybe we will need some sanitization?

sstrasser-globality:
> > @german-glob The idea would be something like this:
> > ```
> > for thing in better_order_by:
> >             if thing.order_dir == OrderDirection.DESC:
> >                 query = query.order_by(desc(text(thing.order_field.value)))
> >             else:
> >                 query = query.order_by(asc(text(thing.order_field.value)))
> > ```
> >   
> > I'm sure there's cleanup I could do on the query and use get_attr or something else fancier, but you get the general idea. https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.order_by
> 
> Yeah, I'm asking because of the SQL injection. We need to make sure it won't open doors for it. Let's check if order by is sent as param to the DB. Maybe we will need some sanitization?

Sanitization happens when the field is mapped to an enum. Ie, if you try to pass in `DROP%20TABLE%20company%3B`, it will try and map that to one of the defined enum fields. If it doesn't map, then it fails deserialization.

sstrasser-globality:
@german-glob (and anyone watching at home) I came up with a fanciey way then using text.Basically, create your enums as follows:

```
class DBTableColumn:

    def __init__(self, column):
        self.column = column


@unique
class OrderBy(BaseEnum):
    FIELD1 = DBTableColumn(TableModel.field1)
    FIELD2 = DBTableColumn(TableModel.field2)
 ```
 The extra class is only needed to play nicer with the SQLAlchemy models/columns. Without it, the enums break and I haven't figured out why yet. 
 
 That then allows you to do:
 ```
        if better_order_by:
            for thing in better_order_by:
                if thing.order_dir == OrderDirection.DESC:
                    query = query.order_by(desc(thing.order_field.value.column))
                else:
                    query = query.order_by(asc(thing.order_field.value.column))
 ```

PR #360, opened 2025-04-22
Remove deprecated `distutils`

salmanmashayekh:
Removing the deprecated `distutils` from the repo.


microcosm_flask/conventions/landing.py
```diff
@@ -29,12 +27,9 @@ def get_properties_and_version():
...
         try:
-            distribution = get_distribution(graph.metadata.name)
-            metadata_str = distribution.get_metadata(distribution.PKG_INFO)
-            package_info = dist.DistributionMetadata()
-            package_info.read_pkg_file(StringIO(metadata_str))
-            return package_info
-        except DistributionNotFound:
+            package_metadata = metadata(graph.metadata.name)
+            return package_metadata
+        except PackageNotFoundError:
```
adamhadani, line 32: Nice @salmanmashayekh looks like almost 1-to-1 replacement

microcosm_flask/converters.py
```diff
@@ -11,3 +11,17 @@ def configure_uuid(graph):
 
     """
     return FlaskUUID(graph.flask)
+
+
+def str_to_bool(value):
+    """
+    Convert a string value to boolean.
+    Similar to distutils.util.strtobool but returns a boolean instead of an int.
```
adamhadani: I'm a little worried about the `similar to .. but returns a boolean instead of an int` - mostly because `microcosm-flask` is used everywhere and this could have some side-effects that will be tricky to debug / will only show up in a while from now when someone pulls in latest microcosm-flask.

So if you want to go with this I would test thoroughly with at least 1-2 services that it works as expected

adamhadani (approved):
LGTM but with important comment / caveat on the `str_to_bool` function and backward compatibility.

salmanmashayekh:
updated the function to match the return type @adamhadani

lfmoisa-globality (approved):

adamhadani (approved):
LGTM, I would try this out locally with 1-2 services that use it, e.g. by `pip install -e .` this package into the venv of a service and run it locally

microcosm_flask/swagger/parameters/__init__.py
```diff
@@ -1,6 +1,6 @@
 from collections.abc import Mapping
 from functools import lru_cache
-from pkg_resources import iter_entry_points
+from importlib.metadata import entry_points
```
adamhadani, line 3: now using `importlib` instead of `pkg_resources`, this was last remaining references to this now deprecated package.

setup.cfg
```diff
@@ -9,7 +9,6 @@ force_grid_wrap = 4
 float_to_top = True
 include_trailing_comma = True
 known_first_party = microcosm_flask
-extra_standard_library = pkg_resources
```
adamhadani, line 12: no longer needed now that we switched to `importlib`

setup.py
```diff
@@ -28,7 +28,7 @@
         "jsonschema>=3.2.0",
         "marshmallow>=3.0.0",
         "microcosm>=4.0.0",
-        "microcosm-logging>=2.0.0",
+        "microcosm-logging>=2.1.0",
```
adamhadani, line 31: once [this](https://github.com/globality-corp/microcosm-logging/pull/65) is merged and released this version will be available - should fix remaining test-failures in CI for this PR cc/ @salmanmashayekh @lfmoisa-globality

setup.py
```diff
@@ -53,9 +53,9 @@
         "lint": [
             "mypy",
             "flake8",
-            "flake8-print",
-            "flake8-logging-format>=1.0.0",
             "flake8-isort",
+            "flake8-logging-format>=1.0.0",
+            "flake8-print",
```
adamhadani, line 58: just sorting deps

